PR #20, opened 2024-03-20
Revert "AME buff"

DangerRevolution:
This reverts commit e271a5da59284954600fa174e96451340a806f86.

# Description

The AME buff made by Wizden is a bandaid solution to Engineers refusing / being inept at / getting killed while setting up the round's power source, that being Singulo or AME or Tesla. As a result, the station quickly runs out of power and everyone suffers as their AME rooms are quite small, due to the philosophy of AME being a placeholder.

As a result, they cranked up the AME power so it can handle the station while Engineers set up power.

Delta stations, and quite literally any non-Wizden station don't give the antimatter engine a broom closet to be set up in so this is never an issue. As a result, this update does nothing for us but make AME so OP that other power is unnecessary.
## TODO:
Wait for:
- [ ] https://github.com/space-wizards/space-station-14/pull/25898

# Changelog
:cl: DangerRevolution
- tweak: AME is now back to it's Delta-V power values

### DIFF
--- a/Content.Server/Ame/AmeNodeGroup.cs
+++ b/Content.Server/Ame/AmeNodeGroup.cs
@@ -182,7 +182,7 @@ public sealed class AmeNodeGroup : BaseNodeGroup
         // Fuel is squared so more fuel vastly increases power and efficiency
         // We divide by the number of cores so a larger AME is less efficient at the same fuel settings
         // this results in all AMEs having the same efficiency at the same fuel-per-core setting
-        return 2000000f * fuel * fuel / cores;
+        return 20000f * fuel * fuel / cores; // Delt V - Revert upstream buff for normal AME operation
     }
 
     public int GetTotalStability()


DebugOk:
> The AME buff made by Wizden is a bandaid solution to Engineers refusing / being inept at / getting killed while setting up the round's power source,

This is just wrong. This was because theres issues with battery values and its waiting on tweaks that are currently pending PRs. Reverting this change without fixing the main power issue will just force engineers to use power sources that'll cause all the lights to flicker.

DebugOk:
> > > The AME buff made by Wizden is a bandaid solution to Engineers refusing / being inept at / getting killed while setting up the round's power source,
> > 
> > 
> > This is just wrong. This was because theres issues with battery values and its waiting on tweaks that are currently pending PRs. Reverting this change without fixing the main power issue will just force engineers to use power sources that'll cause all the lights to flicker.
> 
> Is it this one? The one that fixes battery issues? [space-wizards/space-station-14#25898](https://github.com/space-wizards/space-station-14/pull/25898)

One of them yes. I dont think anything has been done for all the other values that are fucked

VMSolidus:
So I think I can add some context here, and why this doesn't actually matter for DeltaV outside of one very specific map that has an easily fixed issue. The underlying reason why this change was made was for one very specific Wizden map, called Core. Core, famously has a Singularity engine in the exact center of the station. The underlying reason this change was done was that engineers were consistently failing to setup the singularity correctly, in the time limit alotted to them by the station's **intentionally undersized AME**. However, because the entire station was made around Engineering, it would take dozens of hours of mapping work just to expand the AME room to include a fullsize AME, because literally the entire map would need to be altered to accomodate such a change. 

Yes, there is also a bug with SMES. If the total amount of SMES outputs is not great enough to supply the Load required by the station, it causes a brownout issue where the lights flicker. This bug is actually present on Edge station, and is fairly easily fixed with some slight modification to wiring. However that is also itself a mapper issue, not a code issue. For the most part the DeltaV maps did not actually require this change, since we did not have the underlying map issue that Wizden had that justified it. Our maps were designed from the get go with a full size AME, every one of them. 

Undoing this change is justified, since it was irrelevant to the DV codebase in the first place. And it's irrelevant to this codebase as well.

Pspritechologist:
It sounds to me like this is a quick fix to a genuine problem that only cropped up in specific circumstances. If we don't foresee those circumstances arising and we feel the fix causes issues elsewhere, it is perfectly valid to undo it. We should, however, keep in the mind the *real* underlying problem that lead to this fix in the first place and remember that it needs to be dealt with.
The problem isn't false just because we don't expect it to affect us, but it's also perfectly fair to undo this change. Perhaps make a note about it somewhere easy to find and a comment in the YAML regarding the value.

PR #40, opened 2024-03-23
Partial Atmos Refactor

stellar-novas:
# Description

Ports https://github.com/space-wizards/space-station-14/pull/22521. Depends on #39. List of changes is copied below,

  - The MolesArchived float[] field is now read-only
      - It simply gets zeroed whenever the GasMixture is set to null instead of constantly reallocating
  - Airtight query information is now cached in TileAtmosphere
      - This means that it should only iterate over anchored entities once per update.
      - Previously an invalidated atmos tile would cause ProcessRevalidate() to query airtight entities on the same tile six times by calling a combination of GridIsTileAirBlocked(), NeedsVacuumFixing(), and GridIsTileAirBlocked(). So this should help significantly reduce component lookups & entity enumeration.
      - This does change some behaviour. In particular blocked directions are now only updated if the tile was invalidated prior to the current atmos-update, and will only ever be updated once per atmos-update.
      - AFAIK this only has an effect if the invalid tile processing is deferred over multiple ticks, and I don't think it should cause any issues?
  - Fixes a potential bug, where tiles might not dispose of their excited group if their direction flags changed.
  - MapAtmosphereComponent.Mixture is now always immutable and no longer nullable
      - I'm not sure why the mixture was nullable before? AFAICT the component is meaningless if its null?
      - Space "gas" was always immutable, but there was nothing that required planet atmospheres to be immutable. Requiring that it be immutable gets rid of the constant gas mixture cloning.
      - I don't know if there was a reason for why they weren't immutable to begin with.
  - Fixes lungs removing too much air from a gas mixture, resulting in negative moles.
  - GasMixture.Moles is now [Access] restricted to the atmosphere system.
      - This is to prevent people from improperly modifying the gas mixtures (e.g., lungs), or accidentally modifying immutable mixtures.
  - Fixes an issue where non-grid atmosphere tiles would fail to update their adjacent tiles, resulting in null reference exception spam
      - Fixes https://github.com/space-wizards/space-station-14/issues/21732
      - Fixed https://github.com/space-wizards/space-station-14/issues/21210 (probably)
  - Disconnected atmosphere tiles, i.e., tiles that aren't on or adjacent to a grid tile, will now get removed from the tile set. Previously the tile set would just always increase, with tiles never getting removed.
  - Removes various redundant component and tile-definition queries.
  - Removes some method events in favour of just using methods.
  - Map-exposded tiles now get updated when a map's atmosphere changes (or the grid moves across maps).
  - Adds a setmapatmos command for adding map-wide atmospheres.
  - Fixed (non-planet) map atmospheres rendering over grids.

---

# TODO

- [ ] Ensure bird lungs aren't broken

---

# Changelog

:cl: ElectroJr
- fix: Fixed a bug where partially airtight entities (e.g., thin windows or doors) could let air leak out into space


VMSolidus (commented):
This isn't my full approval, I'm going to take a look at why the code is failing first, and then apply some fixes. It should not be that hard to fix, there is most likely a missing Using in WaterVaporReaction according to the test failures. Either there are changes to WaterVaporReaction.cs that we are missing, or this was an error made upstream. Regardless I should be able to fix this.

I'll submit my own approval once I've finished my in-game testing.

stellar-novas:
This is partially fucked https://github.com/space-wizards/space-station-14/issues/26513

DangerRevolution (commented):
is this something already mirrored from Delta PR's?

stellar-novas:
> is this something already mirrored from Delta PR's?
Yes

PR #167, opened 2024-04-21
Mirror: Remove recycling for Syndicate implanters

SimpleStation14:
###### `a8e1489ddb56450cb3a689b95c4c1b407fe3c136`

PR opened by <a href="https://github.com/SlamBamActionman"> SlamBamActionman</a> at 2024-03-12 17:58:45 UTC
PR merged by <a href="https://github.com/web-flow"> web-flow</a> at 2024-03-13 02:02:36 UTC

---

PR changed 1 files with 2 additions and 0 deletions.

- No C#

---

> 
> 
> 
> ## About the PR
> 
> 
> This PR removes the Trash tag for Syndicate implanters, stopping them from being recycled. This means they can no longer be destroyed with the station's Recycler/Material Reclaimer. Non-Syndicate implanters can still be recycled as normal.
> 
> ## Why / Balance
> 
> 
> Being able to recycle implanters completely trivializes the disposal of evidence; if a Material Reclaimer is mapped on the station, it's literally a single button click to fully remove any signs of a contraband having been used, and a Recycler isn't much harder to get going. 
> 
> We already have other systems in the game to hide and alter evidence, such as soap (cleans fingerprints/DNA), stashing in plants, using a storage implant, throwing out of an airlock or just hiding it in a remote place in maintenance. All those methods have some sort of downside or counterplay the antag needs to consider and work around. However, since the recycling is simple, non-suspicious, easily accessible and leaves no trace there is no point in engaging with those other systems as the recycling is always the optimal choice.
> 
> This change is only limited to Syndicate implanters. Non-Syndicate implanters (e.g. bike horn) can be acquired in much larger qualities and thus serve a need to be properly disposed of, in additional to being visually distinct. 
> 
> ## Technical details
> 
> 
> The base implanter prototype has only the Trash tag, so for the base Syndicate implanter the .yaml just sets its tags to blank.
> 
> ## Media
> 
> 
> 
> ## Breaking changes
> 
> 
> **Changelog**
> 
> 
> :cl:
> - tweak: Syndicate implanters can no longer be recycled.
> 

</details>

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Misc/implanters.yml
+++ b/Resources/Prototypes/Entities/Objects/Misc/implanters.yml
@@ -119,6 +119,8 @@
           implantOnly:
             True: {state: broken}
             False: {state: implanter1}
+    - type: Tag
+      tags: []
 
 #Fun implanters
 


DangerRevolution (approved):
removes a cheap bypass

LovelyLophi:
I personally think near all things should be recyclable but should be routed through cargo so it can be sifted through like in ss13.

VMSolidus:
I also have mixed feelings on this, but eh it's fine.

VMSolidus (approved):

DangerRevolution:
> I personally think near all things should be recyclable but should be routed through cargo so it can be sifted through like in ss13.

this would be cool for map based changes :O

PR #199, opened 2024-04-22
Mirror: Democratize C-20r SMG Magazines

SimpleStation14:
###### `c07213c2473453179f525956ff16631a5b77c9cc`

PR opened by <a href="https://github.com/Ghagliiarghii"> Ghagliiarghii</a> at 2024-03-13 18:08:33 UTC

---

PR changed 1 files with 0 additions and 5 deletions.

- No C#

---

> Previously the C-20r magazines were only available for nukies to purchase, but regular syndicate agents could purchase the actual C-20r bundle to get the gun and some magazines. This strikes me an inconsistent. 
> 
> This would be a very minor / QoL buff to normal syndicate agents, as before they could purchase three instances of the .35 auto pistol magazines (for the Viper) and manually load all that ammo into an empty C-20r magazine. This is less annoying to do but would cost 2 TC instead of the 3 TC from the 3 1 TC cost Viper magazines; potentially allowing an agent to have one additional TC for this specific circumstance. It should be noted that this still costs 19 TC (as opposed to the previous 20) so it still doesn't leave a lot of room for other purchases normally.
> 
> I think a 1 TC savings from buying one thing instead of 3 is a reasonable bundle discount.
> 
> 
> 
> 
> ## About the PR
> 
> It doesn't make sense to me that a normal syndicate agent is allowed access to the C-20r magazines through a bundle they can purchase but not access to the magazines by themselves. Additionally it doesn't strike me as a big power difference so I see no reason why they shouldn't be able to purchase these balance wise.
> 
> ## Why / Balance
> 
> This is a slight QoL buff to syndicate agents, it would, at worst, allow an agent one additional TC of purchasing power and a slightly faster reload to a gun that already cost almost all their TC.
> 
> ## Technical details
> 
> Removed whitelist flag for NukeOps for the C-20r magazine in the uplink.
> 
> ## Media
> 
> 
> - [ ] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase
> 
> ## Breaking changes
> 
> 
> **Changelog**
> 
> 
> 
> 

</details>

### DIFF
--- a/Resources/Prototypes/Catalog/uplink_catalog.yml
+++ b/Resources/Prototypes/Catalog/uplink_catalog.yml
@@ -308,11 +308,6 @@
     Telecrystal: 2
   categories:
   - UplinkAmmo
-  conditions:
-  - !type:StoreWhitelistCondition
-    whitelist:
-      tags:
-      - NukeOpsUplink
 
 # For the Cobra
 - type: listing


DangerRevolution (commented):
eh, might as well accept this as I'm sure someone at some point will overhaul weapons & who gets what

DangerRevolution (approved):

LovelyLophi:
Yeah this seemed dumb to me, either syndies should not have the nukie gun or they should have the magazines for it.

DangerRevolution:
> Yeah this seemed dumb to me, either syndies should not have the nukie gun or they should have the magazines for it.

yeah, this is weird but not horrific so might as well merge for someone to overwrite later

VMSolidus (approved):

PR #201, opened 2024-04-22
Mirror: Make radio jammer block suit sensors

SimpleStation14:
###### `dce24dfd03b3ddfe1044297edf9d35bc9f75c523`

PR opened by <a href="https://github.com/nikthechampiongr"> nikthechampiongr</a> at 2024-03-12 17:11:18 UTC

---

PR changed 5 files with 36 additions and 5 deletions.

- Status: Needs Review

---

> 
> 
> 
> ## About the PR
> 
> Radio jammers will now additionally block suit sensors.
> 
> When someone with suit sensors is in range of an active radio jammer, their suit sensor output will be jammed. This will lead to them appearing as frozen on any crew monitors for roughly 10 seconds before they completely disappear from it.
> 
> ## Why / Balance
> 
> 
> This gives syndies more stealth options when doing their objectives, and provides an interesting use for the radio jammer.
> 
> Now on the implementation details:
> 
> With this current implementation you drop off suit sensors after roughly 10 seconds of being jammed. This means you get a 10 second period before anyone actively watching crew monitors can even suspect something's wrong. The reason I have it like this is: 
> a) It's the simplest way to implement this feature.
> b) It's more realistic while still providing a pretty significant advantage.
> 
> Alternatively this could be changed to make the suit sensors submit outdated info indefinitely until the jammer has stopped.
> 
> ## Technical details
> 
> Added a new event which the suit sensors system calls before actually constructing the message it will send to the crew monitoring server. If it is cancelled then the message is not sent. I also intentionally made it so this is checked after a new update time has been set, but if it's a problem it can just be moved above it.
> 
> 
> ## Media
> 
> 
> 
> https://github.com/space-wizards/space-station-14/assets/32041239/3bc815ac-5ec1-4ef1-8594-b390da657d29
> 
> **Changelog**
> 
> 
> 
> :cl:
> - tweak: The radio jammer is now able to jam the signal of suit sensors.

</details>

### DIFF
--- a/Content.Server/Medical/SuitSensors/SuitSensorComponent.cs
+++ b/Content.Server/Medical/SuitSensors/SuitSensorComponent.cs
@@ -87,3 +87,9 @@ public sealed partial class SuitSensorComponent : Component
     [DataField, ViewVariables]
     public bool PreviousControlsLocked = false;
 }
+
+[ByRefEvent]
+public record struct SuitSensorsSendAttemptEvent
+{
+    public bool Cancelled;
+};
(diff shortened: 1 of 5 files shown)


DangerRevolution (approved):

VMSolidus (approved):
Neat!

PR #272, opened 2024-04-22
Mirror: Simplify DoAfterArgs behavior for movement and distance checks

SimpleStation14:
###### `362d56981fd7654f6ebec36f16cc8fabdd04b6a7`

PR opened by <a href="https://github.com/nikthechampiongr"> nikthechampiongr</a> at 2024-02-14 14:26:34 UTC

---

PR changed 75 files with 158 additions and 229 deletions.

- Needs Discussion
- Status: Needs Review

---

> fixes #25113
> 
> 
> 
> ## About the PR
> 
> This pr makes it so that doafter checks for movement uses a single threshold, and with 2 bools to control whether it is checked and whether weightless movement is ignored.
> 
> It also makes doafters checks whether someone can actually interact with whatever they are trying to interact with.
> 
> ## Why / Balance
> 
> I was asked to make these changes to the doafter system on discord.
> 
> I could only test this by running 2 clients at the same time, and stunning one while going on the other to try to do stuff like cuff them, heal them, etc. Obviously this testing is not very extensive but I can't do much locally. If this gets merged we can just adjust the value if needed if it's too forgiving, or too harsh.
> 
> ## Technical details
> 
> Whether a doafter breaks on movement is now decided by BreakOnMove for both the target(if any) and the user.
> If BreakOnWeightlessMove is set to false(true by default) then this check will be skipped when the user is weightless.
> 
> The shareddoaftersystem uses InRangeUnobstructed to determine whether the distance between the user and the target or between the user and their tool is acceptable.
> 
> It should be noted that some systems took advantage of the ability to determine whether to check either the user or the target for movement. This is no longer possible under this pr.
> 
> ## Breaking Changes
> All Doafters now check for distance by default.
> 
> The default movement threshold is now 0.3f instead of 0.1f
> 
> ## Media
> 
> 
> 
> **Changelog**
> 
> 
> 
> no cl no fun

</details>

### DIFF
--- a/Content.Server/Access/Systems/AccessOverriderSystem.cs
+++ b/Content.Server/Access/Systems/AccessOverriderSystem.cs
@@ -55,8 +55,7 @@ public sealed class AccessOverriderSystem : SharedAccessOverriderSystem
 
         var doAfterEventArgs = new DoAfterArgs(EntityManager, args.User, component.DoAfter, new AccessOverriderDoAfterEvent(), uid, target: args.Target, used: uid)
         {
-            BreakOnTargetMove = true,
-            BreakOnUserMove = true,
+            BreakOnMove = true,
             BreakOnDamage = true,
             NeedHand = true,
         };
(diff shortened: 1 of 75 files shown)


DangerRevolution (approved):

VMSolidus:
Actually we CAN'T take this PR. It is fundamentally incompatible with psionic systems, and the in progress psionic rework. This is a strictly negative code PR.

PR #312, opened 2024-04-22
Mirror: Partial atmos refactor

SimpleStation14:
###### `18a35e7e83b2b71ee84b054d44d9ed5e595dd618`

PR opened by <a href="https://github.com/ElectroJr"> ElectroJr</a> at 2023-12-15 03:45:42 UTC

---

PR changed 43 files with 891 additions and 635 deletions.

- Status: Needs Review

---

> This PR reworks how some parts of atmos code work. Originally it was just meant to be a performance and bugfix PR, but it has ballooned in scope. I'm not sure about some of my changes largely because I'm not sure if some things were an oversight or an intentional decision for some reason.
> 
> List of changes:
> - The `MolesArchived float[]` field is now read-only
>   - It simply gets zeroed whenever the `GasMixture` is set to null instead of constantly reallocating
> - Airtight query information is now cached in `TileAtmosphere`
>   - This means that it should only iterate over anchored entities once per update.
>   - Previously an invalidated atmos tile would cause `ProcessRevalidate()` to query airtight entities on the same tile six times by calling a combination of `GridIsTileAirBlocked()`, `NeedsVacuumFixing()`, and `GridIsTileAirBlocked()`. So this should help significantly reduce component lookups & entity enumeration.
>   - This does change some behaviour. In particular blocked directions are now only updated if the tile was invalidated prior to the current atmos-update, and will only ever be updated once per atmos-update. 
>   - AFAIK this only has an effect if the invalid tile processing is deferred over multiple ticks, and I don't think it should cause any issues?
> - Fixes a potential bug, where tiles might not dispose of their excited group if their direction flags changed.
> - `MapAtmosphereComponent.Mixture` is now always immutable and no longer nullable
>   - I'm not sure why the mixture was nullable before? AFAICT the component is meaningless if its null?
>   - Space "gas" was always immutable, but there was nothing that required planet atmospheres to be immutable. Requiring that it be immutable gets rid of the constant gas mixture cloning.
>   - I don't know if there was a reason for why they weren't immutable to begin with.
> - Fixes lungs removing too much air from a gas mixture, resulting in negative moles.
> - `GasMixture.Moles` is now `[Access]` restricted to the atmosphere system.
>   - This is to prevent people from improperly modifying the gas mixtures (e.g., lungs), or accidentally modifying immutable mixtures.
> - Fixes an issue where non-grid atmosphere tiles would fail to update their adjacent tiles, resulting in null reference exception spam
>   - Fixes #21732
>   - Fixes #21210 (probably) 
> - Disconnected atmosphere tiles, i.e., tiles that aren't on or adjacent to a grid tile, will now get removed from the tile set. Previously the tile set would just always increase, with tiles never getting removed.
> - Removes various redundant component and tile-definition queries.
> - Removes some method events in favour of just using methods.
> - Map-exposded tiles now get updated when a map's atmosphere changes (or the grid moves across maps).
> - Adds a `setmapatmos` command for adding map-wide atmospheres.
> - Fixed (non-planet) map atmospheres rendering over grids.
> 
> ## Media
> 
> This PR also includes changes to the atmos debug overlay, though I've also split that off into a separate PR to make reviewing easier (#22520).
> 
> Below is a video showing that atmos still seems to work, and that trimming of disconnected tiles works:
> 
> https://github.com/space-wizards/space-station-14/assets/60421075/4da46992-19e6-4354-8ecd-3cd67be4d0ed
> 
> For comparison, here is a video showing how current master works (disconnected tiles never get removed):
> 
> https://github.com/space-wizards/space-station-14/assets/60421075/54590777-e11c-41dc-b49d-fd7e53bfeed7
> 
> :cl:
> - fix: Fixed a bug where partially airtight entities (e.g., thin windows or doors) could let air leak out into space.
> 

</details>


DangerRevolution (commented):
i'm sure VM can review this

VMSolidus:
I had to double check that my own rework to Space Wind was still compatible with this, and luckily it is since I touched a function that isn't touched by this PR. The two of these combined will hopefully be an extremely significant improvement in atmos performance.

VMSolidus (approved):
This code is quite well made. It still isn't a theoretically perfect atmos, but it this PR represents several substantial improvements to performance into an otherwise very poorly performing system. I am excited to eventually try running this with my Space Wind Rework.

stellar-novas:
This PR introduced a bug that was patched in https://github.com/space-wizards/space-station-14/pull/26441. We should either modify this PR to include the fix, or merge the fix PR at the same time, so downstreams don't get stuck with bugged atmos for a few days.

VMSolidus:
Yep, thanks for the reminder about this. Let's not merge this until I also am ready to immediately afterwards merge https://github.com/space-wizards/space-station-14/pull/26441

VMSolidus:
I need this merged before I can finish the extremely urgent engine PR. I'll include the fixes for this in the engine PR... 
https://github.com/Simple-Station/Einstein-Engines/pull/416

DEATHB4DEFEAT (approved):
Don't have the time to look over this in detail, but it looks fine.

PR #342, opened 2024-04-29
Physics Based Air Throws

VMSolidus:
# Description

I've made it so that when a room is explosively depressurized(or when a body of high pressure air flows into one of lower pressure air), that entities inside are launched by the air pressure with effects according to their mass. An entity's mass is now used as an innate resistance to forced movement by airflow, and more massive entities are both less likely to be launched, and will launch less far than others. While lighter entities are launched far more easily, and will shoot off into space quite quickly! Spacing departments has never been so exciting! This can be made extraordinarily fun if more objects are given the ability to injure people when colliding with them at high speeds.

As a note, Humans are very unlikely to be sucked into space at a typical force generated from a 101kpa room venting into 0kpa, unless they happened to be standing right next to the opening to space when it was created. The same cannot be said for "Lighter-Than-Human" species such as Felinids and Harpies. I guess that's just the price to pay for being cute. :)

On a plus side, because the math behind this is simplified even further than it was before, this actually runs more efficiently than the previous system. 

# TODO

Nothing, this is basically done. I've spent a good 6 hours straight finely tuning the system until I was satisfied that it reflects something close to reality.

# MEDIA

**Before the Space Wind Rework:**

https://github.com/Simple-Station/Einstein-Engines/assets/16548818/0bf56c50-58e6-4aef-97f8-027fbe62331e

**With this Rework:**

https://github.com/Simple-Station/Einstein-Engines/assets/16548818/6be507a9-e9de-4bb8-9d46-8b7c83ed5f1d

# Changelog

:cl: VMSolidus
- add: Atmospheric "Throws" are now calculated using object mass, and behave accordingly. Tiny objects will shoot out of rooms quite fast!


Content.Server/Atmos/EntitySystems/AtmosphereSystem.HighPressureDelta.cs
```diff
@@ -188,7 +184,6 @@ private void ConsiderPressureDifference(GridAtmosphereComponent gridAtmosphere,
             int cycle,
             float pressureDifference,
             AtmosDirection direction,
-            float pressureResistanceProbDelta,
```
Pspritechologist, line 191: I would be willing to bet not a single person has ever called this function on a downstream of Wizden, but I thought I would note that this is technically a public API change, and the kind of thing we'd be documenting in the future for downstreams.

Content.Server/Atmos/EntitySystems/AtmosphereSystem.HighPressureDelta.cs
```diff
@@ -201,37 +196,14 @@ private void ConsiderPressureDifference(GridAtmosphereComponent gridAtmosphere,
...
                 {
-                    var moveForce = maxForce;
-                    moveForce /= (throwTarget != EntityCoordinates.Invalid) ? SpaceWindPressureForceDivisorThrow : SpaceWindPressureForceDivisorPush;
-                    moveForce *= MathHelper.Clamp(moveProb, 0, 100);
-
-                    // Apply a sanity clamp to prevent being thrown through objects.
-                    var maxSafeForceForObject = SpaceWindMaxVelocity * physics.Mass;
-                    moveForce = MathF.Min(moveForce, maxSafeForceForObject);
-
+                    AddMobMovedByPressure(uid, component, physics);
```
Pspritechologist, line 206: This is now technically being called on things that *aren't* mobs, which is counter intuitive.

Content.Server/Atmos/EntitySystems/AtmosphereSystem.HighPressureDelta.cs
```diff
@@ -201,37 +196,14 @@ private void ConsiderPressureDifference(GridAtmosphereComponent gridAtmosphere,
...
 
-                if (maxForce > MovedByPressureComponent.ThrowForce)
+                if (moveForce > physics.Mass)
                 {
-                    var moveForce = maxForce;
-                    moveForce /= (throwTarget != EntityCoordinates.Invalid) ? SpaceWindPressureForceDivisorThrow : SpaceWindPressureForceDivisorPush;
-                    moveForce *= MathHelper.Clamp(moveProb, 0, 100);
-
-                    // Apply a sanity clamp to prevent being thrown through objects.
-                    var maxSafeForceForObject = SpaceWindMaxVelocity * physics.Mass;
```
Pspritechologist, line 232: The old system uses SpaceWindMaxVelocity as a max, which seems to be getting ignored here.

Content.Server/Atmos/EntitySystems/AtmosphereSystem.HighPressureDelta.cs
```diff
@@ -201,37 +196,14 @@ private void ConsiderPressureDifference(GridAtmosphereComponent gridAtmosphere,
...
-                || (physics.BodyType == BodyType.Static && (maxForce >= (component.MoveResist * MovedByPressureComponent.MoveForceForcePushRatio))))
+            if (physics.BodyType != BodyType.Static && !float.IsPositiveInfinity(component.MoveResist))
             {
-                if (HasComp<MobStateComponent>(uid))
-                {
-                    AddMobMovedByPressure(uid, component, physics);
-                }
+                var moveForce = pressureDifference / physics.Mass;
 
-                if (maxForce > MovedByPressureComponent.ThrowForce)
```
Pspritechologist, line 225: 'ThrowForce' also seems to be ignored now. What's that variable for? Is this fine?

Pspritechologist (changes requested):
Certainly very impressive, and I like the effect a lot. I am concerned by the sheer number of variables being discarded (I didn't even note them all!). Even if we feel they aren't required here, they also may just be unused now, and modifying them may not have the implied effect.

Any idea why the items on the table to the left of the breach in the video seem to be unaffected, while things much further away in the room are blasted?

Finally, could we get a before clip recorded of vaguely the same view?

Content.Server/Atmos/EntitySystems/AtmosphereSystem.HighPressureDelta.cs
VMSolidus: ThrowForce was a Constant set at 100f. physics.Mass is approximately 80f for a standard human spessman with all his job equipment on, maybe more depending on what department. I changed this function essentially so that an entity can only be moved by airflow if the force being applied to them is greater than the square of that entity's mass.

Content.Server/Atmos/EntitySystems/AtmosphereSystem.HighPressureDelta.cs
VMSolidus: Eagle has volunteered to test this system for me today on his server, so I'm going to be getting some data from him with it in a live environment. In my own 6 hours of testing, I wasn't able to replicate Wizden's original issue of entities clipping into walls, at least under relatively tame atmospheric circumstances(Tame by my standards meant at most 5000kpa). That's not to say they won't clip if the force is sufficiently high. I'm also waiting for feedback from the original maintainer for this code on what the speed limit is. Which I should actually clamp down by the ApplyLinearImpulse.

VMSolidus:
> Certainly very impressive, and I like the effect a lot. I am concerned by the sheer number of variables being discarded (I didn't even note them all!). Even if we feel they aren't required here, they also may just be unused now, and modifying them may not have the implied effect.
> 
> Any idea why the items on the table to the left of the breach in the video seem to be unaffected, while things much further away in the room are blasted?

That one's an interesting thing to look at. Since objects can only be moved if the force of the air flow is equal to the square of their mass, objects that are lightweight are pulled from farther away(If you look at the top right, you can see that the pens from the Dev medbay were transported all the way from medical to the hole), while objects that are heavier need to be relatively close. The force being applied scales inversely with distance, and the size of the hole in question has an effect on how much force can be applied. The C4 made a very large hole, which in turn spread the force out over a much wider area. The heavier steel on the left wasn't affected because the force wasn't great enough. 

I've also noticed that the relative geometry of the hallways leading up to the hole have their own effects on the airflow.

VMSolidus:
image

I haven't actually removed the variables from the system, reasoning that someone may inevitably try to use them. But they were all either Constants or CVars, not really true math variables. I'm actually not modifying any variables that get passed back, just usingh what information was already available to the function via its access to PhysicsComponent. Another function elsewhere already derives the force of a pressure differential. I did however need to keep some of the exit conditions, because technically the equation I have replaced this with is a parametric equation, it must have some form of limit. 

image

Actually, now that you mention it, because I am dividing by true variables and not constants, I now need to go back and address potential divide by zero exits.

VMSolidus:
https://github.com/Simple-Station/Einstein-Engines/assets/16548818/75f74969-5da0-4bf9-8bdf-70ce158b8b9e


https://github.com/Simple-Station/Einstein-Engines/assets/16548818/23de7a87-dcd7-48b8-9768-b7c20f9666d5

VMSolidus:
Psprite don't merge this yet, I have a pretty good idea how I'm going to reimplement the speed clamp. It basically just means I need to test constants. In ApplyLinearImpulse(uid, vector, physicscomponent), vector is given in the form r(costheta*i + sintheta*j), and in this case dirVec is a vector based on an r=1 circle, and moveForce is my value for r. I simply need to replace these two uses of moveForce with Math.Clamp(moveForce, 0, constant)

And then brute force test constants in a dev environment until I find one where it's impossible to fling a sheet of paper through a steel wall.

rosieposieeee:
hahahahahaha this fucking rules, oh my god

VMSolidus:
Okay! I've added the clamp back in, and did some rather extreme limit testing to determine that this actually is a good number to set the velocity limit to, which is to say it cannot exceed 30 meters per second of acceleration.

VMSolidus:
This is now ready for some excellent reviews.

DangerRevolution (commented):
pog

VMSolidus:
While examining the equations, I figured out another small optimization. Physics components contain an Inverse Mass, equal to 1/Mass. Since multiplication runs faster than division, I can make a small optimization to the equation by multiplying by InvMass instead of dividing by Mass, achieving the same result faster. Since this is a physics system that runs *hundreds* of times per tick, every small improvement matters a lot. 

Getting rid of the division sign means its also no longer possible to divide by zero, so we no longer need to check if mass is 0 to prevent division by zero errors.

Content.Server/Atmos/EntitySystems/AtmosphereSystem.HighPressureDelta.cs
```diff
@@ -183,12 +179,13 @@ private void ConsiderPressureDifference(GridAtmosphereComponent gridAtmosphere,
             tile.PressureDirection = differenceDirection;
         }
 
+        //The EE version of this function drops pressureResistanceProbDelta, since it's not needed. If you are for whatever reason calling this function
+        //And it isn't working, you've probably still got the ResistancePobDelta line included.
```
DEATHB4DEFEAT, line 183:
```suggestion
        //And it isn't working, you've probably still got the pressureResistanceProbDelta line included.
```

Content.Server/Atmos/EntitySystems/AtmosphereSystem.HighPressureDelta.cs
```diff
@@ -201,50 +198,40 @@ private void ConsiderPressureDifference(GridAtmosphereComponent gridAtmosphere,
...
+            // EXPLANATION:
+            // pressureDifference = Force of Air Flow on a given tile
+            // physics.Mass = Mass of the object potentially being thrown
+            // physics.InvMass = 1 divided by said Mass. More CPU efficient way to do division.
+            //
+            // Objects can only be thrown if the force of air flow is greater than the SQUARE of their mass or {SpaceWindMinimumMassThreshold}, whichever is heavier
+            // This means that the heavier an object is, the exponentially more force is required to move it
+            // The force of a throw is equal to the force of air pressure, divided by an object's mass. So not only are heavier objects
+            // less likely to be thrown, they are also harder to throw.
+            // While lighter objects are yeeted easily, and from great distance.
```
DEATHB4DEFEAT, line 210:
```suggestion
            // less likely to be thrown, they are also harder to throw,
            // while lighter objects are yeeted easily, and from great distance.
```

DEATHB4DEFEAT (commented):
I can't really say much but this is cool

VMSolidus:
I've added a Before and After video demonstration to this PR, showing what exactly has been changed.

Content.Server/Atmos/EntitySystems/AtmosphereSystem.HighPressureDelta.cs
```diff
@@ -183,12 +179,13 @@ private void ConsiderPressureDifference(GridAtmosphereComponent gridAtmosphere,
             tile.PressureDirection = differenceDirection;
         }
 
+        //The EE version of this function drops pressureResistanceProbDelta, since it's not needed. If you are for whatever reason calling this function
+        //And it isn't working, you've probably still got the pressureResistanceProbDelta line included.
```
Pspritechologist, line 183:
```suggestion
        //INFO: The EE version of this function drops pressureResistanceProbDelta, since it's not needed. If you are for whatever reason calling this function
        //INFO: And it isn't working, you've probably still got the pressureResistanceProbDelta line included.
```
Some editor configs will highlight this text differently.

Content.Server/Atmos/EntitySystems/AtmosphereSystem.HighPressureDelta.cs
```diff
@@ -201,50 +198,40 @@ private void ConsiderPressureDifference(GridAtmosphereComponent gridAtmosphere,
...
+            // physics.InvMass = 1 divided by said Mass. More CPU efficient way to do division.
+            //
+            // Objects can only be thrown if the force of air flow is greater than the SQUARE of their mass or {SpaceWindMinimumMassThreshold}, whichever is heavier
+            // This means that the heavier an object is, the exponentially more force is required to move it
+            // The force of a throw is equal to the force of air pressure, divided by an object's mass. So not only are heavier objects
+            // less likely to be thrown, they are also harder to throw,
+            // while lighter objects are yeeted easily, and from great distance.
+            //
+            // For a human sized entity with a standard weight of 80kg and a spacing between a hard vacuum and a room pressurized at 101kpa,
+            // The human shall only be moved if he is either very close to the hole, or is standing in a region of high airflow
```
Pspritechologist, line 213:
```suggestion
            //INFO: EXPLANATION:
            // pressureDifference = Force of Air Flow on a given tile
            // physics.Mass = Mass of the object potentially being thrown
            // physics.InvMass = 1 divided by said Mass. More CPU efficient way to do division.
            //
            // Objects can only be thrown if the force of air flow is greater than the SQUARE of their mass or {SpaceWindMinimumMassThreshold}, whichever is heavier
            // This means that the heavier an object is, the exponentially more force is required to move it
            // The force of a throw is equal to the force of air pressure, divided by an object's mass. So not only are heavier objects
            // less likely to be thrown, they are also harder to throw,
            // while lighter objects are yeeted easily, and from great distance.
            //
            // For a human sized entity with a standard weight of 80kg and a spacing between a hard vacuum and a room pressurized at 101kpa,
            // The human shall only be moved if he is either very close to the hole, or is standing in a region of high airflow
```
Some of these should be XML docs for variables on the method itself and not just a comment in the code.
Almost *all* of the can just be in the summary or notes section of the XML docs.

Content.Shared/CCVar/CCVars.cs
```diff
@@ -1049,6 +1049,17 @@ public sealed class CCVars : CVars
...
+        /// <summary>
+        ///     If an object's mass is below this number, then this number is used in place of mass to determine whether air pressure can throw an object.
+        ///     This has nothing to do with throwing force, only acting as a way of reducing the odds of tiny 5 gram objects from being yeeted by people's breath
+        /// </summary>
+        /// <remarks>
+        ///     If you are reading this because you want to change it, consider looking into why almost every item in the game weighs only 5 grams
+        ///     And maybe do your part to fix that? :)
+        /// </remarks>
+        public static readonly CVarDef<float> SpaceWindMinimumMassThreshold =
+            CVarDef.Create("atmos.space_wind_minimum_mass_threshold", 10f, CVar.SERVERONLY);
```
Pspritechologist, line 1061: This isn't a threshold, it's a minimum calculated mass. Clarify the name as such

Content.Server/Atmos/EntitySystems/AtmosphereSystem.HighPressureDelta.cs
```diff
@@ -201,50 +198,40 @@ private void ConsiderPressureDifference(GridAtmosphereComponent gridAtmosphere,
...
+                && !float.IsPositiveInfinity(component.MoveResist))
             {
-                if (HasComp<MobStateComponent>(uid))
-                {
-                    AddMobMovedByPressure(uid, component, physics);
-                }
+                var moveForce = pressureDifference * physics.InvMass;
 
-                if (maxForce > MovedByPressureComponent.ThrowForce)
+                if (moveForce > MathF.MaxMagnitude(physics.Mass, SpaceWindMinimumMassThreshold))
```
Pspritechologist, line 219: Is there a reason `MaxMagnitude` is being used here, rather than simply getting the larger of the numbers?

Content.Server/Atmos/EntitySystems/AtmosphereSystem.HighPressureDelta.cs
```diff
@@ -201,50 +198,40 @@ private void ConsiderPressureDifference(GridAtmosphereComponent gridAtmosphere,
...
                     var dirVec = (direction.ToAngle() + gridWorldRotation).ToWorldVec();
+                    var maxSafeForceForObject = SpaceWindMaxVelocity * physics.Mass;
 
-                    // TODO: Technically these directions won't be correct but uhh I'm just here for optimisations buddy not to fix my old bugs.
+                    // TODO: Consider replacing throw target with proper trigonometry angles.
                     if (throwTarget != EntityCoordinates.Invalid)
                     {
                         var pos = ((throwTarget.ToMap(EntityManager).Position - xform.WorldPosition).Normalized() + dirVec).Normalized();
-                        _physics.ApplyLinearImpulse(uid, pos * moveForce, body: physics);
+                        _physics.ApplyLinearImpulse(uid, pos * Math.Clamp(moveForce, 0, maxSafeForceForObject), body: physics);
```
Pspritechologist, line 230: Do this math outside of this function call and give the variable a name to make the code more readable, and clarify what's happening.

Pspritechologist (dismissed):
A solution needs to be thought of to prevent human-mobs from being untouched. The realism of it is quite unimportant- people need to be able to get flung into space.

This could be making the amount that mass impacts diminishing so that very heavy things can still be flung, or it could be giving mobs Components to change how they're impacted by depressurization.

Content.Server/Atmos/EntitySystems/AtmosphereSystem.HighPressureDelta.cs
VMSolidus: But.. That IS how you get the larger of the two numbers!?!?

VMSolidus:
> A solution needs to be thought of to prevent human-mobs from being untouched. The realism of it is quite unimportant- people need to be able to get flung into space.
> 
> This could be making the amount that mass impacts diminishing so that very heavy things can still be flung, or it could be giving mobs Components to change how they're impacted by depressurization.

I will outright refuse this Psprite because that is extremely, ridiculously out of scope for this PR, because it would necessitate completely rewriting how air flow is calculated in the first place. The original system literally just generated a random number and then used that random number to decide if a human was to be thrown. As a stopgap for the equations for air flow simply being not up to par.

VMSolidus:
> A solution needs to be thought of to prevent human-mobs from being untouched. The realism of it is quite unimportant- people need to be able to get flung into space.
> 
> This could be making the amount that mass impacts diminishing so that very heavy things can still be flung, or it could be giving mobs Components to change how they're impacted by depressurization.

Additionally, I attempted solutions in testing for creating a stopgap to throw humans more consistently, and it had run into many of the same problems the original creator did, where if even the tiniest of player character throws happened, it made the game feel extremely awful to play due to the system *constantly* wresting control from the player. Fixing that is also out of scope for this PR.

VMSolidus:
> A solution needs to be thought of to prevent human-mobs from being untouched. The realism of it is quite unimportant- people need to be able to get flung into space.
> 
> This could be making the amount that mass impacts diminishing so that very heavy things can still be flung, or it could be giving mobs Components to change how they're impacted by depressurization.

So I'll, give it an *attempt* to find a solution to this, but I'll preface it with this. This is an extremely tick intense physics system. This function I optimized runs thousands of times per tick, and every individual line added to it runs up the frametime on an already incredibly frametime hungry system. Adding unnecessary component checks won't do, adding even more math checks here won't do. That improvement you're asking for is going to take making a significant improvement elsewhere, namely in airflow. Airflow pathfinding is extremely bad, creating unusually deep pockets of ultra-fast moving air, while vast portions of the station don't receive anywhere close to the pressure differential that physics dictates they should. For one airflow is only orthogonal adjacent. If I can free it up to perform diagonal movement too, that should in turn dramatically improve the ability of it to throw objects in unusual places, or for heavier objects. That would also allow air to throw objects diagonally, instead of just orthogonal.

For me to accommodate this specific request, I am not going to find a satisfactory solution for it in the function this PR touches.

DEATHB4DEFEAT:
> This function I optimized runs thousands of times per tick

That should not happen for ***anything***, no wonder it sucks for performance

VMSolidus:
> > This function I optimized runs thousands of times per tick
> 
> That should not happen for _**anything**_, no wonder it sucks for performance

I managed to find a extremely common exit condition that bypasses almost 200 lines of script. It should run about 1/100th as often now.

Content.Server/Atmos/EntitySystems/AtmosphereSystem.HighPressureDelta.cs
```diff
@@ -91,6 +90,9 @@ private void AddMobMovedByPressure(EntityUid uid, MovedByPressureComponent compo
 
         private void HighPressureMovements(Entity<GridAtmosphereComponent> gridAtmosphere, TileAtmosphere tile, EntityQuery<PhysicsComponent> bodies, EntityQuery<TransformComponent> xforms, EntityQuery<MovedByPressureComponent> pressureQuery, EntityQuery<MetaDataComponent> metas)
         {
+            //!MAGIC EXIT CONDITION THAT MAKES ALMOST 200 LINES RUN 1/100TH AS OFTEN.
```
DEATHB4DEFEAT, line 93:
```suggestion
```

Content.Server/Atmos/EntitySystems/AtmosphereSystem.HighPressureDelta.cs
```diff
@@ -120,7 +122,8 @@ private void HighPressureMovements(Entity<GridAtmosphereComponent> gridAtmospher
             var gridWorldRotation = xforms.GetComponent(gridAtmosphere).WorldRotation;
 
             // If we're using monstermos, smooth out the yeet direction to follow the flow
-            if (MonstermosEqualization)
+            //WTF:This is bad, don't run this. It just makes the throws worse by somehow rounding them to orthogonal
```
DEATHB4DEFEAT, line 125:
```suggestion
            //TODO This is bad, don't run this. It just makes the throws worse by somehow rounding them to orthogonal
```

Content.Server/Atmos/EntitySystems/AtmosphereSystem.HighPressureDelta.cs
```diff
@@ -181,12 +184,37 @@ private void ConsiderPressureDifference(GridAtmosphereComponent gridAtmosphere,
             tile.PressureDirection = differenceDirection;
         }
 
+        //INFO:The EE version of this function drops pressureResistanceProbDelta, since it's not needed. If you are for whatever reason calling this function
```
DEATHB4DEFEAT, line 187:
```suggestion
        //INFO The EE version of this function drops pressureResistanceProbDelta, since it's not needed. If you are for whatever reason calling this function
```

Content.Server/Atmos/EntitySystems/AtmosphereSystem.HighPressureDelta.cs
```diff
@@ -181,12 +184,37 @@ private void ConsiderPressureDifference(GridAtmosphereComponent gridAtmosphere,
...
+        //INFO:And it isn't working, you've probably still got the pressureResistanceProbDelta line included.
```
DEATHB4DEFEAT, line 188:
```suggestion
        //INFO And if it isn't working, you've probably still got the pressureResistanceProbDelta line included.
```

Content.Server/Atmos/EntitySystems/AtmosphereSystem.HighPressureDelta.cs
```diff
@@ -181,12 +184,37 @@ private void ConsiderPressureDifference(GridAtmosphereComponent gridAtmosphere,
...
+        /// The human shall only be moved if he is either very close to the hole, or is standing in a region of high airflow
+        /// </notes>
+        /// <param name="ent"></param>
+        /// <param name="cycle"></param>
+        /// <param name="pressureDifference"></param>
+        /// <param name="direction"></param>
+        /// <param name="throwTarget"></param>
+        /// <param name="gridWorldRotation"></param>
+        /// <param name="xform"></param>
+        /// <param name="physics"></param>
```
DEATHB4DEFEAT, line 211:
```suggestion
```

Content.Server/Atmos/EntitySystems/AtmosphereSystem.HighPressureDelta.cs
```diff
@@ -199,50 +227,28 @@ private void ConsiderPressureDifference(GridAtmosphereComponent gridAtmosphere,
...
-                    // Apply a sanity clamp to prevent being thrown through objects.
                     var maxSafeForceForObject = SpaceWindMaxVelocity * physics.Mass;
                     moveForce = MathF.Min(moveForce, maxSafeForceForObject);
-
+                    AddMovedByPressure(uid, component, physics);
                     // Grid-rotation adjusted direction
                     var dirVec = (direction.ToAngle() + gridWorldRotation).ToWorldVec();
 
-                    // TODO: Technically these directions won't be correct but uhh I'm just here for optimisations buddy not to fix my old bugs.
+                    // TODO: Consider replacing throw target with proper trigonometry angles.
```
DEATHB4DEFEAT, line 243:
```suggestion
                    //TODO Consider replacing throw target with proper trigonometry angles.
```

Content.Server/Atmos/EntitySystems/AtmosphereSystem.Monstermos.cs
```diff
@@ -676,15 +679,17 @@ private void AdjustEqMovement(TileAtmosphere tile, AtmosDirection direction, flo
...
+            var tileref = tileRef.Tile;
 
-            var chance = MathHelper.Clamp(0.01f + (sum / SpacingMaxWind) * 0.3f, 0.003f, 0.3f);
-
-            if (sum > 20 && _robustRandom.Prob(chance))
+            var tileDef = (ContentTileDefinition) _tileDefinitionManager[tileref.TypeId];
+            if (!tileDef.Reinforced && tileDef.TileRipResistance < delta * MonstermosRipTilesPressureOffset)
+            {
                 PryTile(mapGrid, tile.GridIndices);
+            }
```
DEATHB4DEFEAT, line 692:
```suggestion
                PryTile(mapGrid, tile.GridIndices);
```

Content.Shared/CCVar/CCVars.cs
```diff
@@ -1098,6 +1098,25 @@ public sealed class CCVars : CVars
...
+        ///     If you are reading this because you want to change it, consider looking into why almost every item in the game weighs only 5 grams
+        ///     And maybe do your part to fix that? :)
+        /// </remarks>
+        public static readonly CVarDef<float> SpaceWindMinimumCalculatedMass =
+            CVarDef.Create("atmos.space_wind_minimum_calculated_mass", 10f, CVar.SERVERONLY);
+
+        /// <summary>
+        /// Calculated as 1/Mass, where Mass is the physics.Mass of the desired threshold.
+        /// If an object's inverse mass is lower than this, it is capped at this. Basically, an upper limit to how heavy an object can be
+        /// before it stops resisting space wind more.
```
DEATHB4DEFEAT, line 1115:
```suggestion
        /// 	Calculated as 1/Mass, where Mass is the physics.Mass of the desired threshold.
        /// 	If an object's inverse mass is lower than this, it is capped at this. Basically, an upper limit to how heavy an object can be before it stops resisting space wind more.
```

Content.Shared/CCVar/CCVars.cs
```diff
@@ -1119,7 +1138,21 @@ public sealed class CCVars : CVars
 		///     Also looks weird on slow spacing for unrelated reasons. If you do want to enable this, you should probably turn on instaspacing.
         /// </summary>
         public static readonly CVarDef<bool> MonstermosRipTiles =
-            CVarDef.Create("atmos.monstermos_rip_tiles", false, CVar.SERVERONLY);
+            CVarDef.Create("atmos.monstermos_rip_tiles", true, CVar.SERVERONLY);
+
+        /// <summary>
+        ///     Taken as the cube of a tile's mass, this acts as a minimum threshold of mass for which air pressure calculates whether or not to rip a tile from the floor
+        ///     This should be set by default to the cube of the game's lowest mass tile as defined in .yml prototypes, but can be increased for server performance reasons
```
DEATHB4DEFEAT, line 1145:
```suggestion
        ///     This should be set by default to the cube of the game's lowest mass tile as defined in their prototypes, but can be increased for server performance reasons
```

Content.Shared/CCVar/CCVars.cs
```diff
@@ -1119,7 +1138,21 @@ public sealed class CCVars : CVars
...
+        /// </summary>
+        public static readonly CVarDef<float> MonstermosRipTilesMinimumPressure =
+            CVarDef.Create("atmos.monstermos_rip_tiles_min_pressure", 7500f, CVar.SERVERONLY);
+
+        /// <summary>
+        ///     Taken after the minimum pressure is checked, the effective pressure is multiplied by this amount. This allows server hosts to
+        ///     finely tune how likely floor tiles are to be ripped apart by air pressure
```
DEATHB4DEFEAT, line 1152:
```suggestion
        ///     Taken after the minimum pressure is checked, the effective pressure is multiplied by this amount.
        ///		This allows server hosts to finely tune how likely floor tiles are to be ripped apart by air pressure
```

Content.Shared/Maps/ContentTileDefinition.cs
```diff
@@ -121,5 +121,8 @@ public void AssignTileId(ushort id)
         {
             TileId = id;
         }
+
+        [DataField] public bool Reinforced = false;
+        [DataField] public float TileRipResistance = 125f;
```
DEATHB4DEFEAT, line 126:
```suggestion
        [DataField]
        public bool Reinforced = false;

        [DataField]
        public float TileRipResistance = 125f;
```

DEATHB4DEFEAT (changes requested):

VMSolidus:
This is ready for review. @DEATHB4DEFEAT @Pspritechologist @DangerRevolution @Peptide90 @OldDanceJacket

Aidenkrz:
I believe when we last talked about this, the maxvelocity cvar wasn't working correctly, and you were going to reimplement it after changing up some of the math as it currently does nothing.

VMSolidus:
> I believe when we last talked about this, the maxvelocity cvar wasn't working correctly, and you were going to reimplement it after changing up some of the math as it currently does nothing.

Thanks for the reminder, re implementing the max velocity was actually pretty easy. 
image

VMSolidus:
> I believe when we last talked about this, the maxvelocity cvar wasn't working correctly, and you were going to reimplement it after changing up some of the math as it currently does nothing.

Actually, do you mind passing me your improvements to TileRipping? I'd like to bring them here if they work well enough there.

Aidenkrz:
> Actually, do you mind passing me your improvements to TileRipping? I'd like to bring them here if they work well enough there.

Already deleted the branch, it was pretty terrible code that I just threw together in five minutes for a proof of concept.
IIRC essentially all i did was add a counter that ticked up from the start of it being called, multiply it by a random float, and then multiply the delta by that.

Peptide90 (approved):

DangerRevolution:
Going to play around with this pre-review; while I *do* like the DS14 version, still worth testing. This should be held off from merging until @DEATHB4DEFEAT and/or @Pspritechologist review it too due to the importance/scale :)

DangerRevolution (approved):

DEATHB4DEFEAT (approved):

PR #353, opened 2024-05-01
Make Debris Actually Dangerous

VMSolidus:
# Description

Meant to be a counterpart to this PR: https://github.com/Simple-Station/Einstein-Engines/pull/342

This makes it so that glass shards and floor tiles will embed in people when colliding with them, dealing a small amount of damage. This is also helpful for both making Monstermos with Physics Based Air Throws both less dangerous, and more interesting. Now if an explosion turns a room into a whirlwind of scrap metal, people will actually be significantly concerned about trying not to get hit by debris. 

# Changelog

:cl:
- add: Glass/Crystal shards and floor tiles now deal damage and embed in victims when colliding with them.


DEATHB4DEFEAT:
https://titlecaseconverter.com

VMSolidus:
> https://titlecaseconverter.com

That would be a hilarious thing to make a bot to automate, sending this to Solidus every time he doesn't title case

VMSolidus:
This also prevents floor tiles from just killing EVERYONE in a room, to be fair. Monstermos turns spaced rooms into a very brief blender when combined with Physics Based Air Throws, enough that if a Nukie spaces a room, they better stand aside and wait a brief moment for the shrapnel to fly past. But killing everyone with spammed hits isn't intended

DangerRevolution (approved):
seems interesting, worth getting feedback on how this'll work out

DEATHB4DEFEAT (approved):

PR #365, opened 2024-05-03
Mirror: Store keybind priority

SimpleStation14:
###### `f7a1ffd0aab25d70ac185753d596469fc1e87480`

PR opened by <a href="https://github.com/wrexbe"> wrexbe</a> at 2024-03-23 02:49:07 UTC

---

PR changed 2 files with 2 additions and 1 deletions.

- Changes: UI

---

> Changed it so the priority isn't lost when you set a binding in the UI.
> Also added a priority to MoveStoredItem so it doesn't conflict with Use.
> Fixes https://github.com/space-wizards/space-station-14/issues/26142
> Does not fix old keybinds files, so they will need to reset it, and rebind it.
> 
> A better solution might be to change it so the keybinds are always in the order they appear in the default keybinds folder, to prevent the ordering from changing unpredictably based on what the user overrides.

</details>

### DIFF
--- a/Content.Client/Options/UI/Tabs/KeyRebindTab.xaml.cs
+++ b/Content.Client/Options/UI/Tabs/KeyRebindTab.xaml.cs
@@ -403,7 +403,7 @@ namespace Content.Client.Options.UI.Tabs
                 Mod1 = mods[0],
                 Mod2 = mods[1],
                 Mod3 = mods[2],
-                Priority = 0,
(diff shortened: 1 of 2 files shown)


VMSolidus (approved):

FoxxoTrystan (commented):
👍

DangerRevolution (approved):

PR #396, opened 2024-05-06
Getting up to date

Eagle-0:
There were like 50 EE commits that we didn't have (Which we need to test out languages)


stellar-novas:
Indeed

Though it got auto closed for a valid reason, same thing will happen on the syndie station repo.

You can't merge from master, it has to be a different branch.

PR #409, opened 2024-05-16
Listening Post Earlier Start

VMSolidus:
# Description

This was something that always annoyed me back when I made this for DeltaV. One key feature of the Listening Post was that it would have acted to encourage traitors to actually buy the encryption key bundle, since it had the possibility of letting them get radio support from a listening post. At the time Debug insisted it should only ever spawn extremely late in the around, giving it a sufficiently low weight that it would only appear usually an hour into the round. By this point all of the traitors had either been caught or had already done their thing, thus defeating the entire point of it. It also had an absurdly high minimum player count relative to the actual threat the outpost serves, ensuring that it would never appear on lowpop servers. 

Seriously not even stationwide lethal threats such as Xenos Ventcritters had as high of a minimum player count.

# Changelog

:cl:
- tweak: Syndicate Listening Posts can now appear any time in a round, and can appear on low-pop servers.


DEATHB4DEFEAT (approved):

DangerRevolution (approved):

FoxxoTrystan (commented):
👍

PR #411, opened 2024-05-18
Temporary CI/CD Fix

DEATHB4DEFEAT:
# Description

Ports https://github.com/DeltaV-Station/Delta-v/commit/7e3ba621d325db4d838c4d0fb675b5173f3111df until we get the real fix.
Should allow tests to run, though it still fails most of the time according to Null.


DangerRevolution (approved):

VMSolidus (approved):
Send it

PR #433, opened 2024-06-03
Default to Separated UI Layout

DEATHB4DEFEAT:
# Description

<sub>~~Change from Nyano~~</sub>
Uses space a lot better. The old layout still exists if people want it for whatever reason.
I also renamed "Default" to "Overlay", which should stay anyway if the default change is denied.

---

<p>

## Separated

image

## Overlay

image

</p>
</details>

---

# Changelog

:cl:
- tweak: UI layout now defaults to separated, the old one is still available in settings

### DIFF
--- a/Content.Client/Gameplay/GameplayState.cs
+++ b/Content.Client/Gameplay/GameplayState.cs
@@ -93,16 +93,16 @@ namespace Content.Client.Gameplay
             var screenTypeString = _configurationManager.GetCVar(CCVars.UILayout);
             if (!Enum.TryParse(screenTypeString, out ScreenType screenType))
             {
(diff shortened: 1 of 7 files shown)


DangerRevolution (approved):

VMSolidus (changes requested):
Do you mind including screenshots of the two different UIs, so that we can see in the PR the differences visually of what's being changed?

DEATHB4DEFEAT:
> Do you mind including screenshots of the two different UIs, so that we can see in the PR the differences visually of what's being changed?

Done.

VMSolidus (approved):
OH. Yea that's a significant improvement.

PR #459, opened 2024-06-14
Refactor the Language System

Mnemotechnician:
# Description
The language PR was merged early and OH GOD I ALREADY REGRET IT

This PR is intended to provide the missing refactors and address the issues that were missed due to the early merge.

---

# TODO
- [X] Introduced a polymorphic obfuscation property to the LanguagePrototype - now it supports more than just 2 hardcoded methods, and each method can be configured per-language. Currently there are 3 obfuscation methods: replacement (same as replacement accent), obuscation by syllables and obfuscation by phrases.
- [X] Refactored the existing obfuscation methods to not be a big hardcoded mess.
- [X] Updated the existing languages accordingly: animalistic languages are now less of an unreadable mess and include less syllables. Certain languages like binary and snake seriously benefit from that.
- [X] Refactored the existing commands in response to the never-addressed review (it got lost among hundreds of others)
- [X] Refactored the commands to be more user-friendly (you can now use the number of the language in saylang and languageselect which can allow using keybinds to switch between languages)
- [X] Moved a lot of obfuscation-related stuff from server to shared. The actual obfuscation process, however, is still done on the server. That may or may not be subject to change, too.
- [X] Refactored the entire process of resolution of entities' languages. Instead of raising an event every time it's required to learn what languages an entity knows, the lists of ALL languages available to the entity (including via translators) is stored in LanguageSpeakerComponent and only updated when necessary (e.g. when a translator gets toggled). The list of languages the entity knows on its own is now stored in LanguageKnowledgeComponent.
- [X] Made handheld translators automatically change your current language when activated.
- [X] Rewrote the translator implanter system, now using the real implants and implanters
- [ ] Rebalance science stuff (translators are incredibly expensive for what they're worth)
- [ ] Uhhh stuff

---

<p>

N/A for now

</p>
</details>

---

# Changelog

:cl:
- tweak: Translator implants are now proper implants that can be removed.
- tweak: Animalistic languages should now look less messy.
- fix: Hopefully fixed language menu desync and other issues.


Content.Shared/Language/LanguagePrototype.cs
```diff
@@ -10,15 +9,13 @@ public sealed class LanguagePrototype : IPrototype
...
+    [DataField("obfuscation")]
+    public ObfuscationMethod Obfuscation = ObfuscationMethod.Default;
 
     /// <summary>
-    ///     Lists all syllables that are used to obfuscate a message a listener cannot understand if obfuscateSyllables is true.
-    ///     Otherwise uses all possible phrases the creature can make when trying to say anything.
+    ///     A list of replacement phrases used in <see cref="ObfuscationMethod"/>
     /// </summary>
     [DataField(required: true)]
     public List<string> Replacement = [];
```
Pspritechologist, line 21: Now that language obfuscators are custom classes, this should be moved to the two that actually use it, since not all languages will.

Pspritechologist:
Love to see this being done, I was actually just coming now to check on what needs to be fixed with Languages :)

I agree that it was merged prematurely, for some of the pending fixes please take a look at the unresolved review comments on the last pr.
There's some great looking stuff already in here :+1:

Pspritechologist:
> Introduced a polymorphic obfuscation property to the LanguagePrototype

I was about to do this myself! :P Looks great.
> Moved a lot of obfuscation-related stuff from server to shared. The actual obfuscation process, however, is still done on the server. That may or may not be subject to change, too.

Prediction = Good but we *really* don't want obfuscation happening on Shared. The client should *only ever* become aware of the message *post* modifications, otherwise the un-modified data is simply available to them. We need to be aware of vectors for cheating, and exposing unnecessary data to the client is one of the biggest.

Mnemotechnician:
> Prediction = Good but we really don't want obfuscation happening on Shared. The client should only ever become aware of the message post modifications, otherwise the un-modified data is simply available to them. We need to be aware of vectors for cheating, and exposing unnecessary data to the client is one of the biggest.

Indeed, that's what I've been keeping in mind while writing the coode, but the current networking approach is just ever so stupid and bug-prone...

Mnemotechnician:
This is mostly done now I think. I'll leave the balance part out for now, I don't feel like I can come up with something reasonable myself.

VMSolidus:
image

Test fail is legitimate.

Mnemotechnician:
Oopsie.

FoxxoTrystan (commented):
Reviewed and looks fucking amazing, will base future Language updates from it.

EE Maints, please review.

DangerRevolution (approved):

VMSolidus (approved):
Excellent work.

Issue #474, opened 2024-06-18
Audio Sucks


DEATHB4DEFEAT:
# Description

Audio kinda sucks and has for maybe ever.
Especially since the audio rework.
There's a lack of sliders for specific sounds, everything is on `Master` except for a couple of things that don't matter much.
I keep running into the issue of joining a modern codebase after playing old Parkstation and hating how loud and how quiet random things are that are all bunched into `Master` audio, and just end up turning everything down a *ton*.
Even if things weren't weirdly volumed, everything is just *so much louder* between before and after the audio rework, it's annoying reconfiguring my settings between the two.
The issue of playing the different codebases isn't a very big one, and is definitely not one most players will have, especially not for long as old servers die or modernize.

The `Needs Discussion` and `Help Wanted` tags are here for a reason, please comment.

### General issues

- [ ] Event music and announcements are very loud
	- Event music has a toggle, but that's not enough
	- Announcement sound is resolved by https://github.com/Simple-Station/Einstein-Engines/issues/415
- [ ] Admin audio is just a toggle
- [ ] Machine and other ambience not from the newish background ambience system or whatever are only in the `Master` channel but should probably be in their own or the existing `Ambience` one
	- What does the `Ambience Max Sounds` slider do/does it matter or need to exist? And why is it not near the `Ambience` slider?
- [ ] `Lobby & Round-end Music` should just be a slider instead of a toggle, or have them next to each other in a compact way
	- [ ] Same with everything else

### Odd sounds

- [ ] Plastic water bottles shouldn't have the glass sound effect

### Oddly loud/quiet sounds

- FTL end is loud?
- [x] Casing throws are too loud - https://github.com/Simple-Station/Einstein-Engines/issues/478
- [x] https://github.com/Simple-Station/Einstein-Engines/pull/457 - https://github.com/Simple-Station/Einstein-Engines/pull/624

Issue #467, opened 2024-06-16
Physics System Overhaul Tracker


VMSolidus:
# Description

Space Station 14 has a key feature available to it that gives it a significant advantage over coding for Space Station 13, that being a fully fledged Physics System. Under this system, every entitity in the game has access to information such as Mass, Density, Size, Hitboxes, and so on. This information gets referenced by other systems, which allows the game to have a sort of tactile responsiveness to a player's actions. Yet unfortunately, the core game of SS14 doesn't have consistent applications of it. Objects are missing mass, systems that could account for it don't, some systems have incomplete usage of it. 

I have been for awhile seeking to expand upon the game's usage of Physics, and have been pushing for a series of PRs that add complicated interactivity to the game. The start of which is my Space Wind Rework, which has gradually turned into a substantial overhaul of the game's Atmos system. But SWR isn't the whole story, because a whole host of other systems have emergent interactions with Space Wind. 

# Related Issues:

- #460 
- #461 
- #352 
- #337 

# Related PRs:

- #464 
- #458 
- #457 
- #342 
- #341

WarMechanic:
it would be cool if bullets were affected by physics to some degree, such that their trajectory would arc around gravitational singularities or a potential psionic power

DEATHB4DEFEAT:
> it would be cool if bullets were affected by physics to some degree, such that their trajectory would arc around gravitational singularities or a potential psionic power

I believe they are and do.

WarMechanic:
hm, ill have to test this at some point

DocNITE:
It is actual? 👀

Issue #460, opened 2024-06-14
The Great Mass'ening


VMSolidus:
# Description
99% of all in-game objects obtain their Physics Component and Fixtures Component from BaseItem, which has resulted in nearly every object in the entire game sharing the same weight of only 5 kilograms. This game has a surprising variety of physics systems that actually care about the mass of objects, so it's pretty much a crime against humanity that every object has the same mass. 
> Throwing objects involves a set amount of force being applied as an impulse. If every object weighs the same, that means players throw them all with the same force.
> Bullet casings weigh FIVE KILOGRAMS, and are ejected by this same throwing system, so their astoundingly high weight means they go pretty much nowhere. Even the bullet casings should have different weights, so that they can be launched with their own unique velocity. 
> #342 reworks Space Wind to respect an object's mass, but every item having the same mass means they all get yeeted with the same force. This is wildly incorrect.

I would like for most of the game's items to have mass relatively accurately reflect a real world expectation of mass. This is primarily done by adjusting their fixture density. Not every item must be individually set, in most cases there are Parent hierarchies of similar items, such that if all the items that share a Parent are essentially identical to each other in mass, then only their parent needs to have its Mass fixed.

PR #478, opened 2024-06-19
Make Spent Casing Ejection Not Suck

VMSolidus:
# Description

Part of #467, #460, and #474

This is a small PR that corrects a math error in SharedGunSystem, causing shell casings to be "Thrown" directly downwards instead of in an actually cinematic and exciting arc. While I'm at it, I also corrected the fixture of base shell casings to favor "Bounciness", and decreased its mass to approximately 100 grams. Finally, I added a sound for when casings bounce off of walls, which wasn't present before. 

https://github.com/Simple-Station/Einstein-Engines/assets/16548818/56bb4ecc-d5eb-4b36-853b-42f05374150d

:cl:
- fix: Spent bullet casings now fly away from a shooter in a cinematic manner, rather than fall at their feet.


DangerRevolution (approved):
image

Maybe one day wall hitboxes won't suck lol

DEATHB4DEFEAT (approved):

PR #11, opened 2024-03-17
Lamia & Segmented Entity System

VMSolidus:
## About the PR
This is a draft PR for an upcoming new playable species, the Lamia. Lamia are an older species from the 2022 era of Nyanotrasen, and were previously abandoned code that was dropped from the game on October 13th, 2022. I was able to locate what I believe to be the last remaining branch containing Lamia, from a defunct server repository that ran an October 12th, 2022 build of Nyanotrasen. Thus I began a project to refurbish this code for use on modern SS14. 

The Lamia I wish to PR are only recognizable from the original Lamia in that they share the Tail Segment System. That is, they prominently feature a completely unique mechanic whereby their body is composed of multiple entities linked together in a chain. The original version of this code had a great many bugs related to the game's physics system, and it was severely limited by the Robust Toolbox engine at the time. In the year since Lamia's abandonment, Robust Toolbox has gone through many iterations and improvements, and has deprecated large parts of the physics system that old Lamia utilized. 

## Redesigns from the original 2022 Lamia, AKA Nyanotrasen Lamia. Vs. DeltaV Lamia
The original Nyanotrasen Lamia were made with several limitations in mind. Their size was heavily restricted by the physics engine at the time, we aren't beholden to that same limitation anymore. Thus DeltaV Lamia are vastly larger than the Nyanotrasen Lamia, featuring a tail that is 5 tiles in length, with fully functional physics collisions. They were also not able to wear Hardsuits due to limitations of the SpriteComponent, and thus were instead designed around having a "Barotrauma resistance". On DeltaV code, we can arbitrarily state that species use different optional sprites for items, therefore its possible to have for example a Nukie Hardsuit, with its equipped-outerwear state, equipped-outerwear-lamia, equipped-outerwear-lamiainitialsegment, equipped-outerwear-lamiasegment, and so on. The Lamia Segments can simply state that if they equip a hardsuit, they utilize the -lamiasegment sprite option. Therefore its no longer necessary to create an entire new item solely so that snakes can wear a hardsuit. 

**Positive Traits**
- Extreme Size. Lamia are 5 tiles long, and weigh as much as a car. They make a mockery of mass contests, and they can push physics objects around simply by slithering into them.
- Significantly larger health bar. A Lamia is put into critical condition at 200 damage, and dies at 300 damage. This is offset by the Lamia having a certain percentage of damage taken by the tail transferred to the main body.
- High resistance to forced movement. Space Wind at standard pressure cannot move them.
[...]

PR #479, opened 2024-06-19
Cherrypick "Shoot Over Bodies" And Related PRs

VMSolidus:
# Description
This is a manual cherry-pick of the following PRs:
https://github.com/space-wizards/space-station-14/pull/27905
https://github.com/space-wizards/space-station-14/pull/28072
https://github.com/space-wizards/space-station-14/pull/28571

I REQUIRE these for my work in PR #11 , and cannot complete said PR without these cherry-picks. This set of PRs from Wizden adds a feature where entities can selectively opt-out of being shot at unless a player intentionally targets them, which I can use as a simple and elegant solution to one of the largest glaring issues for Segmented Entities. I could simply give Lamia segments the new RequireProjectileTargetComponent, which adds them to the system. Future segmented entities such as the hypothetical "Heretic Worm" may or may not use this feature, depending on their intended implementation.


DangerRevolution:
Copy the bodies over from the previous PR's; so whoever comes looking doesn't have to dive down several links (and because I also don't want to dive down these links)

VMSolidus:
> Copy the bodies over from the previous PR's; so whoever comes looking doesn't have to dive down several links (and because I also don't want to dive down these links)

Yea I just checked, I actually can't do that.

DangerRevolution (approved):

Peptide90 (approved):

PR #484, opened 2024-06-24
Port All Carrying/PseudoItem/EscapeInventory Tweaks From DeltaV

Mnemotechnician:
# Description
This cherry-picks the following two PRs of mine from delta-v:
- https://github.com/DeltaV-Station/Delta-v/pull/1118 (this one is a port of two other frontier PRs, see the original description for details)
- https://github.com/DeltaV-Station/Delta-v/pull/1232

Encompassing a total of 8 distinct changes:
1. Fixes dropping the carried person when walking to a different grid (from station to shuttle or vice versa. Walking into space however will still make you drop them) and also makes sure that pressing shift while being carried does not make you escape.
2. Ensures that the carried person is always centered relative to the parent (under certain conditions, such as walking near a gravitational anomaly, their position can change, and that leads to really weird effects)
3. Fixes the mass contest in CarryingSystem that caused stronger entities to take longer to escape than weaker ones.
4. Adds popups for when you're getting picked up or being stuffed into a bag as a pseudo-item (e.g. a felinid)
5. Adds an action to stop escaping an inventory. This action gets added to your action bar when you attempt escaping and gets removed when you stop or escape. It applies both to carrying and items (hampsters, felinids, whatever else).
6. Adds a sleep action for pseudo-items stuffed inside a suitable bag. The bag must have a special component, which is added to the base backpack item and thus inherited by all soft bags (duffels, satchels, etc). Contrary to a popular belief, sleeping IS PURELY COSMETICAL and does not provide healing. (Beds provide healing when you buckle into them and that healing does not depend on whether or not you're sleeping)
7. Makes it so that when you try to take a pseudo-item out of the bag (e.g. a felinid), you automatically try to carry them (if you don't have enough free hands, they will be dropped on the floor like usually), and enables you to insert the carried person into a bag, but only if they're a pseudo-item (e.g. felinid).
8. Allows pseudoitems to be inserted into bags even when there are other items (as long as there's enough space)

---

## For technical details and video showcases, see the original PRs

This PR is split into separate commits so different parts can be reverted if deemed unneccessary.

---

# Changelog
:cl:
- fix: Carrying is less likely to behave erratically or suddenly interrupt now.
- add: You can now see when someone is trying to pick you up, and also you can interrupt your attempt at escaping from their hands or inventory.
- add: You can now properly take Felinids out of bags and place them inside.
- add: Scientists have discovered that Felinids can sleep in bags.


VMSolidus:
I haven't given this a deeper look than only a cursory glance, but I'm fascinated by the fact that this system for whatever baffling reason has its own private implementation of the original (bad) MassContest code. This is itself terrible, and it's not even the fault of the DeltaV PR, they're just swapping multiply with divide. I'm not going to deny moving it here because of that, but whichever comes first, this or the Reworked Mass Contests, the other is getting changed and I'm yeeting the damn carrying system implementation of Old MassContest.

DangerRevolution (changes requested):
should all be in base game folders with comments removed

Mnemotechnician:
> should all be in base game folders with comments removed

image

DangerRevolution (approved):

VMSolidus (approved):

PR #486, opened 2024-06-24
Fewer Mid-Round Events

DEATHB4DEFEAT:
# Description

Ports and improves https://github.com/Simple-Station/Parkstation-Friendly-Chainsaw/pull/59

Slows rounds down so the station doesn't always have to evacuate in a blaze and can instead leave complete.
Also makes Survival not stupidly destructive and gives a chance for the shift to last a decent time.
Allows servers to configure the times themselves if they want to, so defaults don't matter too much.

---

# Changelog

:cl:
- tweak: Mid-round events will occur much less often

### DIFF
--- a/Content.Server/StationEvents/BasicStationEventSchedulerSystem.cs
+++ b/Content.Server/StationEvents/BasicStationEventSchedulerSystem.cs
@@ -4,7 +4,9 @@ using Content.Server.GameTicking.Rules;
 using Content.Server.GameTicking.Rules.Components;
 using Content.Server.StationEvents.Components;
 using Content.Shared.Administration;
+using Content.Shared.CCVar;
 using JetBrains.Annotations;
+using Robust.Shared.Configuration;
 using Robust.Shared.Random;
 using Robust.Shared.Toolshed;
 using Robust.Shared.Utility;
@@ -20,6 +22,7 @@ namespace Content.Server.StationEvents
     {
         [Dependency] private readonly IRobustRandom _random = default!;
         [Dependency] private readonly EventManagerSystem _event = default!;
+        [Dependency] private readonly IConfigurationManager _config = default!;
 
         protected override void Ended(EntityUid uid, BasicStationEventSchedulerComponent component, GameRuleComponent gameRule,
             GameRuleEndedEvent args)
@@ -57,8 +60,8 @@ namespace Content.Server.StationEvents
         /// </summary>
         private void ResetTimer(BasicStationEventSchedulerComponent component)
         {
-            // 5 - 25 minutes. TG does 3-10 but that's pretty frequent
-            component.TimeUntilNextEvent = _random.Next(300, 1500);
+            component.TimeUntilNextEvent = _random.Next(_config.GetCVar(CCVars.GameEventsBasicMinimumTime),
+                _config.GetCVar(CCVars.GameEventsBasicMaximumTime));
         }
     }
 
--- a/Content.Server/StationEvents/RampingStationEventSchedulerSystem.cs
+++ b/Content.Server/StationEvents/RampingStationEventSchedulerSystem.cs
@@ -70,7 +70,7 @@ public sealed class RampingStationEventSchedulerSystem : GameRuleSystem<RampingS
     {
         var mod = GetChaosModifier(uid, component);
 
-        // 4-12 minutes baseline. Will get faster over time as the chaos mod increases.
-        component.TimeUntilNextEvent = _random.NextFloat(240f / mod, 720f / mod);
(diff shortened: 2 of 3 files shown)


DangerRevolution (approved):

VMSolidus (approved):
I genuinely appreciate that this lets us CVar configure survival rounds.

VMSolidus (changes requested):
Actually, now that I think about it, this system should actually use "Hot Loaded" CVars for the game event timers, which following our repository conventions, would necessitate creating a StationEventSchedulerSystem.CVars.cs file similar to the ones present in AtmosphereSystem.CVars.cs and CPRSystem.CVars.cs. It would be extremely useful for server hosts to be able to on the fly modify the next instance of the event scheduler. Plus it would make all of this code a lot cleaner by removing all of the repeated instances of _config.GetCVar(CCVars.variable).

DEATHB4DEFEAT:
> Actually, now that I think about it, this system should actually use "Hot Loaded" CVars for the game event timers, which following our repository conventions, would necessitate creating a StationEventSchedulerSystem.CVars.cs file similar to the ones present in AtmosphereSystem.CVars.cs and CPRSystem.CVars.cs. It would be extremely useful for server hosts to be able to on the fly modify the next instance of the event scheduler. Plus it would make all of this code a lot cleaner by removing all of the repeated instances of _config.GetCVar(CCVars.variable).

<details><summary>This is horrible, I'm not doing it like this.</summary>
<p>

image

</p>
</details> 

I do have another similar idea, though, one second.

DEATHB4DEFEAT:
> I do have another similar idea, though, one second.

It did not work.

VMSolidus:
> > I do have another similar idea, though, one second.
> 
> It did not work.

I'm not sure why the convention of essentially moving all of the _config.GetCvar stuff to a separate file isn't acceptable? From where I'm standing it looks pretty good? We're essentially just moving all of the Cvar logic to a separate file, so the main bulk of the system's logic becomes a lot more human readable.

DEATHB4DEFEAT:
> I'm not sure why the convention of essentially moving all of the _config.GetCvar stuff to a separate file isn't acceptable? From where I'm standing it looks pretty good? We're essentially just moving all of the Cvar logic to a separate file, so the main bulk of the system's logic becomes a lot more human readable.

There's so much extra stuff for negligible "benefit" that will break easier, and it's very unclear what those variables actually are/do and what you can do with them.

VMSolidus:
> > I'm not sure why the convention of essentially moving all of the _config.GetCvar stuff to a separate file isn't acceptable? From where I'm standing it looks pretty good? We're essentially just moving all of the Cvar logic to a separate file, so the main bulk of the system's logic becomes a lot more human readable.
> 
> There's so much extra stuff for negligible "benefit" that will break easier, and it's very unclear what those variables actually are/do and what you can do with them.

Can you help me understand your point of view here? Can you explain in more detail what it is you are suggesting might break if done this way? Can we not simply put documentation on the variables for the EventSchedulerSystem.CVars.cs file?

DEATHB4DEFEAT:
> Can you explain in more detail what it is you are suggesting might break if done this way?

Anything. Adding more redundant things mean it's more likely someone will forget or not know something, and it won't be changed.

> Can we not simply put documentation on the variables for the EventSchedulerSystem.CVars.cs file?

No, that's more duplicate stuff.
(Inheritdoc does exist though)

VMSolidus:
> > Can you explain in more detail what it is you are suggesting might break if done this way?
> 
> Anything. Adding more redundant things mean it's more likely someone will forget or not know something, and it won't be changed.
> 
> > Can we not simply put documentation on the variables for the EventSchedulerSystem.CVars.cs file?
> 
> No, that's more duplicate stuff. (Inheritdoc does exist though)

Well my underlying problem with this now is that we are dealing with two different standards present in the repository, and no consistent guideline on which to do here. I genuinely could have swore I was under the impression we actually wanted to more consistently apply this specific application of .Component.cs, System.cs, and System.CVars.cs, and that we had discussed this fairly early on in the repository's history. 

Does inheritdoc inherit across different system spaces?

DEATHB4DEFEAT:
> Well my underlying problem with this now is that we are dealing with two different standards present in the repository, and no consistent guideline on which to do here.

What you sent me to is not at all standard and will not be.

> I genuinely could have swore I was under the impression we actually wanted to more consistently apply this specific application of .Component.cs, System.cs, and System.CVars.cs, and that we had discussed this fairly early on in the repository's history.

We do, we just need to figure out a decent way of doing the CVar part.

> Does inheritdoc inherit across different system spaces?

Don't know why it wouldn't.

VMSolidus (approved):
I will keep this in mind in the future.

PR #315, opened 2024-04-22
Mirror: Restrict door remotes to only being able to manipulate doors relevant to their type

SimpleStation14:
###### `93e3aed26e30375058aa44482236410ff31484f0`

PR opened by <a href="https://github.com/nikthechampiongr"> nikthechampiongr</a> at 2024-03-24 00:28:39 UTC

---

PR changed 1 files with 3 additions and 8 deletions.

The PR had the following labels:

---

<details open="true"><summary><h1>Original Body</h1></summary>

> 
> 
> 
> ## About the PR
> 
> This makes it so door remotes can no longer manipulate doors just because their user has access to them. 
> 
> ## Why / Balance
> 
> Being able to bolt doors and put them on emergency access from a distance is extremely powerful. You should not be able to do it just because you have the access in your id card, it should be limited to the capabilities of the door remote.
> 
> It also makes no sense.
> 
> ## Technical details
> 
> Relevant calls to check for access, and try open/bolt doors had their arguments changed to pass in the door remote instead of the user so now only the door remote's accesses are checked.
> 
> ## Media
> 
> 
> - [x] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase
> 
> ## Breaking changes
> 
> 
> **Changelog**
> 
> 
> 
> :cl:
> - tweak: Door remotes can now only control doors specific to their type and do not use the access of their user.

</details>

PR #491, opened 2024-06-28
Revert "Mirror: Restrict Door Remotes to only Being Able to Manipulate Doors Relevant to Their Type"

stellar-novas:
Reverts Simple-Station/Einstein-Engines#315

As [I brought up](https://github.com/Simple-Station/Einstein-Engines/pull/315#issuecomment-2105861135) in the original pr, this change was made to counter power gaming, which isn't as much of an issue on EE downstreams, and adversely affects the engineering department, without any major benefits.

### DIFF
--- a/Content.Server/Remotes/DoorRemoteSystem.cs
+++ b/Content.Server/Remotes/DoorRemoteSystem.cs
@@ -50,8 +50,10 @@ namespace Content.Shared.Remotes
                 return;
             }
 
+            // Holding the door remote grants you access to the relevant doors IN ADDITION to what ever access you had.
+            // This access is enforced in _doorSystem.HasAccess when it calls _accessReaderSystem.IsAllowed
             if (TryComp<AccessReaderComponent>(args.Target, out var accessComponent)
-                && !_doorSystem.HasAccess(args.Target.Value, args.Used, doorComp, accessComponent))
+                && !_doorSystem.HasAccess(args.Target.Value, args.User, doorComp, accessComponent))
             {
                 _doorSystem.Deny(args.Target.Value, doorComp, args.User);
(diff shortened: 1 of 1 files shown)


DangerRevolution:
Happy to merge but; thinking about it that it makes sense that a remote only opens what the remote is for?

stellar-novas:
I think it makes sense for a remote to be able to link up to the user's ID card.

DangerRevolution (commented):
I'm going to be honest, if this gets two reviews and is merged that's okay but I am genuinely against it, I can't really get behind it...

VMSolidus:
So I actually don't really agree with reverting this.

VMSolidus (approved):

Content.Server/Remotes/DoorRemoteSystem.cs
```diff
@@ -61,15 +63,18 @@ private void OnBeforeInteract(Entity<DoorRemoteComponent> entity, ref BeforeRang
             switch (entity.Comp.Mode)
             {
                 case OperatingMode.OpenClose:
-                    if (_doorSystem.TryToggleDoor(args.Target.Value, doorComp, args.Used))
+                    // Note we provide args.User here to TryToggleDoor as the "user"
+                    // This means that the door will look at all access items carryed by the player for access, including
+                    // this remote, but also including anything else they are carrying such as a PDA or ID card.
```
DEATHB4DEFEAT, line 68:
```suggestion
                    // This means that the door will look at all access items carried by the player for access, including
                    //   this remote, but also including anything else they are carrying such as a PDA or ID card.
```

DEATHB4DEFEAT (approved):

PR #492, opened 2024-06-29
Localises 

DangerRevolution:
# Description

Localises da game; opening this PR to run tests on Github and while I do manual checks; I will expand on this section later.

This PR cleans up the salad description; as that causes errors when running a locale script.

This PR cleans up the backpacks.yml folder in the Delta-V back clothing folder; as having a folder with .yml on the end of it caused errors in the script.

Some (many) entities with no description; and no description that can be inherited from a parent instead point to the parent to *look* for descriptions from that parent. This opens about 80 or so non-critical errors when building.

Needs a comb-through as some "entities" that aren't real entities haven't been localised. Such as "stress test" debug entities.

---

# TODO

- [X] Initial Locale
- [ ] Checking for errors
- [ ] Discussing whether or not this is worth it

---


DangerRevolution:
Setting to draft; it looks like I'm going to have to clean up a fair amount of the base entities that were localised.

Mnemotechnician:
AFAIK entity names & descriptions are usually kept in-line so forks in other languages can easily localize them without modifying upstream code, this may disrupt the regular workflow.

DangerRevolution:
> AFAIK entity names & descriptions are usually kept in-line so forks in other languages can easily localize them without modifying upstream code, this may disrupt the regular workflow.

Not sure what you mean by this?

VMSolidus:
Oh hey this is finally a thing! I've been wanting this for awhile, ostensibly so that I can go through the list and make my fun "Lore Through Item Descriptions" thing

VMSolidus:
Any idea what's going on with the errors here? I would gladly approve this actually. But it would have to actually work first.

DEATHB4DEFEAT (changes requested):
Everything is in `prototypes/prototypes`.
Add newlines to the end of every file.
Tons of these are referencing themselves or don't have/need a name/desc.
If you're going to move localization of any prototype to ftl, the fields need to be removed from their prototypes.

I think it'd be better to make (I have) and use a standard format instead of using many different formats everywhere.
Many things around entities (their loc format in this case) are very hard-coded to work only with them, and we need something for everything.

DangerRevolution:
This PR is morbillions away from being done; was meant to be re-drafted. I just don't know how to do it on mobile...

DEATHB4DEFEAT:
Closing due to inactivity, reopen if you ever return to this.

PR #507, opened 2024-07-02
Fix Clothing Quick-Equip

Mnemotechnician:
# Description
For some reason, ClothingSystem.TryEquip would return false if there's a do-after to equip the clothing. This had caused all quick-equip attempts on SMALL ITEMS to fail and ended up with every SMALL clothing item being equipped into one of the pocket slots (which have no equip delays).

Also fixes quick swap - see the comments below.

# Changelog

:cl:
- fix: Equipping clothing using the Z key works correctly again.

### DIFF
--- a/Content.Shared/Clothing/EntitySystems/ClothingSystem.cs
+++ b/Content.Shared/Clothing/EntitySystems/ClothingSystem.cs
@@ -62,6 +62,11 @@ public abstract class ClothingSystem : EntitySystem
     {
         foreach (var slotDef in userEnt.Comp1.Slots)
         {
+            // Do not attempt to quick-equip clothing in pocket slots.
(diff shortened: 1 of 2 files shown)


Mnemotechnician:
Apparently doing the same to TryUnequip fixes quick swap too :trollface:

Mnemotechnician:
Right, I decided to just make the clothing system skip pocket slots while quick equipping. It's actually stupid that pockets were considered valid slots for this action, because for example if you were to hold and ID card and press Z, it would swap the item in your pocket with that ID, which had personally pissed me off before. So this should kill two bugs with one stone.

DangerRevolution (approved):
Hopefully this doesn't have any consequences we can't foresee

VMSolidus (approved):

PR #509, opened 2024-07-04
Restore Old Event Rates

Mnemotechnician:
# Description
This restores the basic/survival event rates to how they were before before #486 (however, it keeps the CVars created in it).

- 5-25 minutes for basic instead of 15-35
- 4-12 minutes for survival instead of 20-45

---

# Why
The PR made it so that survival rounds actually have less events than extended - that is considering that survival rounds are supposed to be troublesome and filled with events while extended ones are supposed to be a chill alternative.

Other forks don't bother with editing CVars every 5 minutes, so this "opportunity to configure their experience" was never addressed. Based on the feedback I've received from other players, the change was rather negative.

From my personal experience, survival rounds became an extended extended, where during a 2-hour shift you can get a grand total of one power outage and two mice infestations, and nothing else at all.

---

# Changelog

:cl:
- tweak: Events should now occur as frequently as before. Note: server owner can configure the frequency on their server manually.

### DIFF
--- a/Content.Shared/CCVar/CCVars.cs
+++ b/Content.Shared/CCVar/CCVars.cs
@@ -112,14 +112,14 @@ namespace Content.Shared.CCVar
         ///     Close to how long you expect a round to last, so you'll probably have to tweak this on downstreams.
         /// </summary>
         public static readonly CVarDef<float>
-            EventsRampingAverageEndTime = CVarDef.Create("events.ramping_average_end_time", 120f, CVar.ARCHIVE | CVar.SERVERONLY);
+            EventsRampingAverageEndTime = CVarDef.Create("events.ramping_average_end_time", 40f, CVar.ARCHIVE | CVar.SERVERONLY);
 
         /// <summary>
         ///     Average ending chaos modifier for the ramping event scheduler.
         ///     Max chaos chosen for a round will deviate from this
         /// </summary>
         public static readonly CVarDef<float>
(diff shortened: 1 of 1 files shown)


VMSolidus (approved):

DEATHB4DEFEAT (changes requested):
I accidentally swapped the initial values in my PR, but survival is supposed to get way more events than extended over the shift, just change its' modifier CVars instead if it doesn't and correct the swap.

Mnemotechnician:
The issue is that even if we swap the values, events will still be extremely rare compared to any other fork.

Mnemotechnician:
I am an active player and I have never heard anyone complain about how there's too many events, ever. However, I did hear others complain that even with the old event rates there were way too little actually dangerous events.

DangerRevolution (approved):
I personally believe that it's better to have events be more frequent; and have forks tune it down rather than vice versa. I also believe it lets EE show off some of it's more mechanical changes.

VMSolidus:
So since our default is MRP, it really wasn't a terrific idea to have the default for other servers to use be that of HRP style event timers. People had a bit of an odd disconnect with it, expecting the event timers to be relatively consistent, but then had servers with extremely uneventful rounds be the default.

Really, long term I still want to do a Director System, but that'll be on hold until after we get our hands on the Antag Refactor(Having ALL antags being a function you can call is incredibly nice for that.)

PR #512, opened 2024-07-05
Heavyweight Drunk Trait + Drunk Traits Rework

WarMechanic:
# Description

Adds the inverse of lightweight drunk, which makes you less susceptible to the effects of ethanol.
(more specifically, it halves the damage of alcohol and you need to drink twice as much to feel the effects in the first place)

To make the change happen, `LightweightDrunk` component was reworked to:
- A) no longer change any drunk effects (including non-alcohol related drunkness like bloodloss)
- B) instead multiply the effects of ethanol in your bloodstream

I chose this route in particular, because the other option of multiplying the amount of ethanol gained from alcohols is nonsense.

---

# TODO

- [X] Add a `TryMetabolizeReagent` event to `MetabolizerSystem.cs` so a `LightweightDrunkSystem.cs` can listen to the event and multiply the effects of "Ethanol" specifically.
- [X] Add the Heavyweight Drunk trait
- ~~Fix a minor spelling mistake that caused the trait condition to not show up~~
- [ ] Would we be able to name trait files after categories? I don't want to put every positive trait in a file named 'skills.yml'

---

<p>

Example Media Embed

</p>
</details>

---

# Changelog

:cl:
- add: Added the Heavyweight Drunk trait, which doubles your alcoholism potential.


WarMechanic:
before i undraft this im actually going to move `LightweightDrunkComponent.cs` to server because the client shouldn't use it anymore

DangerRevolution (approved):
Is this something that's possible to replicate for being high?

VMSolidus:
Had to update it to include the new engine version, this was one I think where I had to very much by hand modify the merge conflict, rather than pick one.

VMSolidus (approved):

PR #546, opened 2024-07-13
Shock Collar Tweak

Mnemotechnician:
# Description
Shock collars were a useful thing in nyano for a long time, but security could never put them to use because they could be removed instantly, so they were forgotten about.

Now that we have enquip delays, shock collars can be useful for controlling unruly prisoners (or unruly secoffs who like to harm baton passengers). Only two problems is there's no way to manufacture them, and they can be quite deadly when used repeatedly. This PR solves both of these issues, and also fixes a little bug:
- Shock collars can be manufactured at the secfab.
- Shock collars deal 1 shock damage instead of 5.
- The shock ignores insulation.

image

</p></details>

---

# Changelog
:cl:
- tweak: Shock collars can now be manufactured in the security techfab. They also deal less damage and ignore insulated gloves.

### DIFF
--- a/Resources/Prototypes/Entities/Structures/Machines/lathe.yml
+++ b/Resources/Prototypes/Entities/Structures/Machines/lathe.yml
@@ -743,6 +743,7 @@
       - ShellShotgunPractice
       - WeaponLaserCarbinePractice
       - WeaponDisablerPractice
+      - ShockCollard
       # DeltaV - .38 special ammo - Add various .38 special ammo to security techfab
       - MagazineBoxSpecial
       - MagazineBoxSpecialPractice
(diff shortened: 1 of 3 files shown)


Resources/Prototypes/Entities/Structures/Machines/lathe.yml
```diff
@@ -743,6 +743,7 @@
       - ShellShotgunPractice
       - WeaponLaserCarbinePractice
       - WeaponDisablerPractice
+      - ShockCollard
```
DangerRevolution, line 746:
```suggestion
      - ShockCollar
```

DangerRevolution (approved):
Interesting to see it had a lathe recipe but was never used.. I wonder why?

VMSolidus (approved):

PR #549, opened 2024-07-14
Oracle Refactor

Mnemotechnician:
# Description
Refactors the oracle system and component, making them more customizable and less trash.

- Most of the nyano shitcode was rewritten in accordance with the new standards and in a less hardcoded manner (all features were preserved, with some changes).
- Replaced the accumulator field pattern with the "nextX" pattern for things oracle does on certain intervals.
- Removed some code duplication and bloat. 
- Gave oracle a 75% chance to request technology and 25% to request a plant.
- When requesting a technology, oracle will only request what's either already researched, or can be researched soon (that is, the prerequisites of the research are complete and the research isn't locked for any reason).
- If there's no research server at the moment when oracle demands an item, it will only demand a plant
- When dispensing rewards, oracle can now spawn 1, 3, or 5 crystals at once, and research disks worth 5k, 10k, and 20k research points. Each one of those have different probabilities (3 crystals and 5k points still prevail, but occasionally you can get the more expensive rewards).
- When dispensing rewards, oracle now throws them at the person who fulfilled the request instead of simply spawning them at their feet.

Almost every oracle thing (except for throwing and the amount of liquid dispensed) can now be configured in the yaml prototype of the oracle.

# TODO
Forget it

https://github.com/user-attachments/assets/9d4be44f-37d5-4072-a6e6-f194764f7ff6

image

</p></details>

---

# Changelog
:cl:
- tweak: Oracle requests are now more likely to be aligned with the current research.


Mnemotechnician:
Test fail is legit, I'mma look into it tomorrow.

Mnemotechnician:
Fixed the test fail by removing the test :trollface: 

In all honesty, I don't think it was ever necessary.

VMSolidus:
Test fails are legitimate, but you can get rid of 2 out of 3 of the test fails by replacing your ComponentInit subscription with MapInitEvent. The reason for fail 2 & 3 is that you aren't actually allowed to modify an entity directly during MapInit(ComponentInit happens during MapInit!), because modifying entities will actually change the map file!

Luckily, MapInitEvent can be treated just like ComponentInit for this purpose, except that it waits for the map to finish loading if it is already loading. Note that despite what the name implies, it *also* works like ComponentInit, in that an entity spawned during a round will still trigger its MapInitEvent subscription.

Mnemotechnician:
I just removed the init event callback. Should work now, I had no idea this was related to the oracle system at all.

DangerRevolution:
do you plan to finish the toDo before you want reviews or is this okay to review?

Content.IntegrationTests/Tests/Nyanotrasen/Oracle/OracleTest.cs
```diff
@@ -38,33 +38,36 @@ public async Task AllOracleItemsCanBeTurnedIn()
...
-                // otherwise this pair cannot be considered clean, and the
-                // CleanReturnAsync call would need to be removed.
-                mapManager.DeleteMap(testMap.MapId);
-            });
+            // await server.WaitAssertion(() =>
+            // {
+            //     var allProtos = oracleSystem.GetAllProtos(oracleComponent);
+            //     var coordinates = testMap.GridCoords;
+            //
+            //     Assert.That((allProtos.Count > 0), "Oracle has no valid prototypes!");
```
VMSolidus, line 49: If this test is no longer necessary at all due to the new way Oracle is coded, then the code for the test should be deleted, rather than commented out. When refactoring, we do not need to keep vestigial remnants of bad code.

VMSolidus (changes requested):

VMSolidus (changes requested):
Since we are also now taking maintainer responsibility for these previously Abandoned code systems, we are assuming code ownership of this system. Therefore per Einstein-Engines' agreement with the original code owner, we are to signify code ownership of these files by moving them out of the Nyanotrasen namespace.

Content.Server/Nyanotrasen/Research/Oracle/OracleSystem.cs
```diff
@@ -1,258 +1,288 @@
...
+    {
+        if (!_solutions.TryGetSolution(oracle.Owner, OracleComponent.SolutionName, out var fountainSol))
+            return;
 
-        var sol = new Solution();
-        var reagent = "";
+        // Why is this hardcoded?
+        var amount = MathF.Round(20 + _random.Next(1, 30) + _glimmer.Glimmer / 10f);
+        var temporarySol = new Solution();
+        var reagent = _protoMan.Index(oracle.Comp.RewardReagents).Pick(_random);
```
VMSolidus, line 170: Maybe you could have the oracle be aware of the research tier of the item being requested, and improve the rewards by some factor of research tier? That way science is given better rewards for harder to obtain items?

Content.Server/Nyanotrasen/Research/Oracle/OracleSystem.cs
```diff
@@ -1,258 +1,288 @@
...
     {
-        if (!_solutionSystem.TryGetSolution(uid, OracleComponent.SolutionName, out var fountainSol))
-            return;
+        foreach (var rewardRandom in oracle.Comp.RewardEntities)
+        {
+            // Spawn each reward next to oracle and throw towards the target
+            var rewardProto = _protoMan.Index(rewardRandom).Pick(_random);
+            var reward = EntityManager.SpawnNextToOrDrop(rewardProto, oracle);
+            _throwing.TryThrow(reward, throwTarget, recoil: false);
+        }
```
VMSolidus, line 157: Same here, Iirc research tier is an integer? You could for instance make it drop one item per integer value of the research tier(or 1, if there was no research tier).

VMSolidus (changes requested):

Content.Server/Nyanotrasen/Research/Oracle/OracleSystem.cs
Mnemotechnician: If you can suggest a particular implementation, sure, otherwise introducing new mechanics is not within the scope of this PR.

VMSolidus: That's fair enough! I personally was planning on making this section a little more variable with a future implementation in refactored psionics, but for now it's good enough.

VMSolidus (approved):

Mnemotechnician:
Whatever I could find, I moved out of nyano folders. Cannot guarantee the server won't crash because of that.

VMSolidus:
> Whatever I could find, I moved out of nyano folders. Cannot guarantee the server won't crash because of that.

We're good here. Luckily, the Linters are designed to catch issues like that. It wouldn't have passed the checks if we were missing a redirected Using, or if a path didn't resolve to the correct location anymore. :)

DangerRevolution (approved):

PR #560, opened 2024-07-18
Mass Contests Affect Camera Shake

VMSolidus:
# Description

I made this entirely on a whim to prove a point in the middle of a conversation. I have not tested this code, but I'm confident that it will work as advertised. 

This refactors the public KickCamera function so that it no longer: Divides by 1, divides by 1 again, multiplies by 1, normalizes a vector(turning its A value to 1), and then divides said vector by 1 again for good measure.

Instead 1 has been replaced by MassContest, wooh! I also got rid of redundant steps. The effect of this is that characters with greater than human-standard mass are slightly more resistant to all sources of camera shake(Explosions, gravity generator turning off, gunfire, being hit by flying objects, etc), while characters with less than standard human mass experience a greater amount of camera shake.

You're welcome. Now go find a Felinid to shake vigorously. 

# Changelog

:cl:
- add: Camera Shake is now affected by character mass. Small characters experience more camera shake. Large characters experience less.

### DIFF
--- a/Content.Client/Camera/CameraRecoilSystem.cs
+++ b/Content.Client/Camera/CameraRecoilSystem.cs
@@ -1,5 +1,4 @@
 using System.Numerics;
-using Content.Client.Overlays;
 using Content.Shared.Camera;
 using Content.Shared.CCVar;
 using Content.Shared.Contests;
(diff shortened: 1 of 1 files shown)


DEATHB4DEFEAT (approved):
You should test this

DangerRevolution (approved):

PR #561, opened 2024-07-18
Remove Ending Punctuation From Emote Detection

DEATHB4DEFEAT:
# Description

Replaces the tons of super inconsistent emote defs with consistent and more dynamic ones.

---

<p>

https://github.com/user-attachments/assets/eed4a1c9-7fe7-45a9-acb9-cbd64dd319e0

</p>
</details>

---

# Changelog

:cl:
- fix: Emotes now consistently work with any or no punctuation


Resources/Prototypes/DeltaV/Voice/speech_emotes.yml
```diff
@@ -2,242 +2,78 @@
 - type: emote
   id: HarpyHonk
   category: Vocal
-  chatMessages: [honks.]
+  chatMessages: [ honks ]
   chatTriggers:
-    - honk
-    - honk.
```
VMSolidus, line 8: Slightly overzealous here. The singular form honk and past-tense honked are still necessary.

Content.Server/Chat/Systems/ChatSystem.Emote.cs
```diff
@@ -150,6 +152,10 @@ public bool TryPlayEmoteSound(EntityUid uid, EmoteSoundsPrototype? proto, string
     private void TryEmoteChatInput(EntityUid uid, string textInput)
     {
         var actionLower = textInput.ToLower();
+        // Replace ending punctuation with nothing
+        if (Punctuation.Any(punctuation => actionLower.EndsWith(punctuation)))
+            actionLower = actionLower.Remove(actionLower.Length - 1);
+
         if (!_wordEmoteDict.TryGetValue(actionLower, out var emote))
             return;
```
VMSolidus, line 160: I hope this isn't *actually* modifying the emotes. Because I would prefer that it doesn't, and that the system simply makes it not consider punctuation when deciding whether or not to play an emote sound.

Second, I'd like to request if possible, consider making it so that it reads the whole emote for a sound effect. I've been annoyed for a very long time that I can't include the word *PEW* in a full sentence emote, and the sound won't play.

Resources/Prototypes/DeltaV/Voice/speech_emotes.yml
```diff
@@ -2,242 +2,78 @@
...
-    - ringing!
-    - ringing?
-    - ringed
-    - ringed.
-    - ringed!
-    - ringed?
-    - rang
-    - rang.
-    - rang!
-    - rang?
```
VMSolidus, line 48: Here too, and in all other places. Don't remove the different tenses of words. It's only necessary to remove punctuation variants.

VMSolidus (changes requested):

Resources/Prototypes/DeltaV/Voice/speech_emotes.yml
DEATHB4DEFEAT: why?

Content.Server/Chat/Systems/ChatSystem.Emote.cs
DEATHB4DEFEAT: > I hope this isn't actually modifying the emotes. Because I would prefer that it doesn't, and that the system simply makes it not consider punctuation when deciding whether or not to play an emote sound.

Look at the video, it doesn't change the text.

> Second, I'd like to request if possible, consider making it so that it reads the whole emote for a sound effect. I've been annoyed for a very long time that I can't include the word PEW in a full sentence emote, and the sound won't play.

Do that in another PR.

VMSolidus (approved):

VMSolidus:
The video was hidden by a tab, and my ADHD suffering self with no adderall today didn't see the tab. My mistake. Carry on.

Content.Server/Chat/Systems/ChatSystem.Emote.cs
```diff
@@ -9,6 +10,7 @@ namespace Content.Server.Chat.Systems;
 public partial class ChatSystem
 {
     private FrozenDictionary<string, EmotePrototype> _wordEmoteDict = FrozenDictionary<string, EmotePrototype>.Empty;
+    private IReadOnlyList<string> Punctuation { get; } = new List<string> { ",", ".", "!", "?", "-", "~", "'", "\"", };
```
Pspritechologist, line 13:
```suggestion
    private IReadOnlyList<string> Punctuation { get; } = [ ",", ".", "!", "?", "-", "~", "'", "\"",  ];
```

Pspritechologist (approved):

PR #563, opened 2024-07-19
Replace Hugs with Looking

leonardo-dabepis:
# Description

Replaces the hugging action with the looking action. So that then when you click on someone, it says that you look at them rather than you hug them, and vice versa. It's awkward to accidentally hug people. Also ODJ told me to put this here instead of my downstream 😤 

---

# TODO

- [ ] See if it actually works (it should in theory)
- [x] Changed the files

---

<p>

Example Media Embed

</p>
</details>

---

# Changelog

:cl:
- tweak: Replaced hugging with looking

### DIFF
--- a/Resources/Locale/en-US/interaction/interaction-popup-component.ftl
+++ b/Resources/Locale/en-US/interaction/interaction-popup-component.ftl
@@ -77,11 +77,11 @@ comp-window-knock = *knock knock*
 
 fence-rattle-success = *rattle*
 
-## Hugging players
+## Looking at players
 
-hugging-success-generic = You hug {THE($target)}.
-hugging-success-generic-others = { CAPITALIZE(THE($user)) } hugs {THE($target)}.
-hugging-success-generic-target = { CAPITALIZE(THE($user)) } hugs you.
+looking-success-generic = You look at {THE($target)}.
+looking-success-generic-others = { CAPITALIZE(THE($user)) } looks at {THE($target)}.
+looking-success-generic-target = { CAPITALIZE(THE($user)) } looks at you.
 
 ## Other
 
--- a/Resources/Prototypes/DeltaV/Entities/Mobs/Player/harpy.yml
+++ b/Resources/Prototypes/DeltaV/Entities/Mobs/Player/harpy.yml
@@ -7,9 +7,9 @@
   - type: CombatMode
   - type: InteractionPopup
     successChance: 1
-    interactSuccessString: hugging-success-generic
+    interactSuccessString: looking-success-generic
     interactSuccessSound: /Audio/Effects/thudswoosh.ogg
-    messagePerceivedByOthers: hugging-success-generic-others
+    messagePerceivedByOthers: looking-success-generic-others
   - type: MindContainer
     showExamineInfo: true
   - type: Input
--- a/Resources/Prototypes/DeltaV/Entities/Mobs/Player/harpy.yml
+++ b/Resources/Prototypes/DeltaV/Entities/Mobs/Player/harpy.yml
@@ -8,7 +8,6 @@
   - type: InteractionPopup
     successChance: 1
     interactSuccessString: looking-success-generic
-    interactSuccessSound: /Audio/Effects/thudswoosh.ogg
     messagePerceivedByOthers: looking-success-generic-others
   - type: MindContainer
     showExamineInfo: true
--- a/Resources/Prototypes/DeltaV/Entities/Mobs/Player/vulpkanin.yml
+++ b/Resources/Prototypes/DeltaV/Entities/Mobs/Player/vulpkanin.yml
@@ -7,9 +7,9 @@
     - type: CombatMode
     - type: InteractionPopup
       successChance: 1
-      interactSuccessString: hugging-success-generic
+      interactSuccessString: looking-success-generic
       interactSuccessSound: /Audio/Effects/thudswoosh.ogg
-      messagePerceivedByOthers: hugging-success-generic-others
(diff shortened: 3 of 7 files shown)


DEATHB4DEFEAT (changes requested):
I'd prefer a tap or something, but this is fine.
You need to remove the sound effect.

leonardo-dabepis:
> I'd prefer a tap or something, but this is fine. You need to remove the sound effect.

hehe i forgor. It's gone now.

WarMechanic:
Do you think you could also increase the interaction range of looking?

leonardo-dabepis:
> Do you think you could also increase the interaction range of looking?

I'm not certain how I'd do that, but that is a good idea.

Resources/Locale/en-US/interaction/interaction-popup-component.ftl
```diff
@@ -77,11 +77,11 @@ comp-window-knock = *knock knock*
...
 
-## Hugging players
+## Looking at players
 
-hugging-success-generic = You hug {THE($target)}.
-hugging-success-generic-others = { CAPITALIZE(THE($user)) } hugs {THE($target)}.
-hugging-success-generic-target = { CAPITALIZE(THE($user)) } hugs you.
+looking-success-generic = You look at {THE($target)}.
+looking-success-generic-others = { CAPITALIZE(THE($user)) } looks at {THE($target)}.
+looking-success-generic-target = { CAPITALIZE(THE($user)) } looks at you.
```
Mnemotechnician, line 84: Do not override those. Create separate entries, but keep the old ones.

Mnemotechnician:
And what if I want to hug a person and not awkwardly stare at them?

DangerRevolution:
> And what if I want to hug a person and not awkwardly stare at them?

you switch to emote and type it out

OldDanceJacket:
> Hugging is also used to make people stand up from a bed. Conventionally this is used to indicate to a patient "you are good to go".
> 
> While hugging also doesn't make _that_ much sense for that purpose, converting this to "looking" seems less so. Have you considered an alternative to this?
> SS14.Loader_QT2PLyiDuJ.mp4

Hugging isn't what's getting them out of the bed, it's unbuckling which existed before hugging was put into the game. A patient will hear the unbuckle sound and stand up all the same(you can also right click and go to unbuckle or drag them off).

I understand that medical is absolutely **_starved_** for mechanics but this ain't it.

stellar-novas:
> Hugging isn't what's getting them out of the bed, it's unbuckling which existed before hugging was put into the game. A patient will hear the unbuckle sound and stand up all the same(you can also right click and go to unbuckle or drag them off).

Related to this, empty hand interact should either just look, or should have a different emote string, since it would be weird if just looking at someone unbuckled them.

OldDanceJacket:
> > Hugging isn't what's getting them out of the bed, it's unbuckling which existed before hugging was put into the game. A patient will hear the unbuckle sound and stand up all the same(you can also right click and go to unbuckle or drag them off).
> 
> Related to this, empty hand interact should either just look, or should have a different emote string, since it would be weird if just looking at someone unbuckled them.
> 
> I'm in favor of the shoulder tap.

Looking at someone isn't what's unbuckling them, it's simply left clicking on the person who's buckled into a chair or bed, this doesn't change the range at where this is possible either so I fail to see how this is 'weird'. If it seems strange to you, stop thinking 'looks at person, they unbuckle' being how it flows. Because this isn't how you seen it previously with hugging it's 'you look at the person, you unbuckle the person' they're separate things. Frankly I find it weird that my doctor is hugging me anyways???
This discussion on this seems to be caught up on a minor point--Medical treating people in beds.

What about literally every other situation, like security buckling someone out of bed/chair with cuffs? Random hug situations happen there, it's the same situation but with different department afterall.

Changing the text from "hugs" to "looks at" just removes the weird behavior associated with having a dedicated hug prompt in the game, where you'll have folks across tables hug you, drive by hugs from mimes, trying to click on someone to attack them, trying to grab an item as someone walks over it, etc., with a far more generic prompt that fills more situations to reduce awkwardness and be less immersion breaking. _(Shoulder tap doesn't do this, I'd even argue shoulder tap would cover less situations than hugs)_

The entire point of this PR is to be a very low scope, simple thing anyways, if you want added functionality or context specific interactions _(I would love open hand on someone who has slipped to help them off the floor for instance)_ this is not within the scope of the PR.

BluCottonCandy:
The idea of "Looks at you" is a lot creepier to me than a hug.

Can't say I like the idea of magically unbuckling someone with my eyes when they're fully healed and need to stand up from the bed.

Mnemotechnician:
I once again remind you that you should leave the old interaction strings as they were and add new ones **below**. Otherwise this will cause issues later.

Resources/Locale/en-US/interaction/interaction-popup-component.ftl
DangerRevolution: This + remove the hugging sound

DangerRevolution (approved):

DangerRevolution (changes requested):
Wrong button, keep the old strings for use by other people and remove the hugging sound effect

OldDanceJacket:
With the inclusion of  #710  this PR no longer serves a function as the hug interaction has simply been outright removed. 
Reasoning for this can be found in  #710 . 

This PR was intended to be a simple temporary fix for a better solution(#710 ) anyways; So I'll be closing it now.

leonardo-dabepis:
oh shit I forgot I made this lol

PR #572, opened 2024-07-20
Fix 125 Toolshed Commands Being Unusable for Anyone without +HOST

Mnemotechnician:
# Description
For reasons unknown, AdminManager has two command permission managers: _commandPermissions and _toolshedCommandPermissions. It used to load normal command permissions into the former, and toolshed command permissions into the latter. Since _toolshedCommandPermissions is NEVER actually used in checking whether a player can execute a command, all toolshed commands remained unavailable to anyone without +HOST.

This PR provides a bandaid fix for that: it makes it so that the same permissions are loaded into both managers at the same time. It's necessary to load them into _commandPermissions in order to allow regular players to execute them, and it's also necessary to load them into _toolshedCommandPermissions because otherwise the ToolshedManager will complain about those commands lacking permission flags.

This should also fixes some commands such as `spawn`, `pos`, `comp` being inaccessible to admins with +DEBUG and more.

image

image

Admin-only commands are unaffected:

image

</p></details>

---

# Changelog

:cl:
- fix: Fixed toolshed command permissions. This will mostly affect admins who don't have full host access.

### DIFF
--- a/Content.Server/Administration/Managers/AdminManager.cs
+++ b/Content.Server/Administration/Managers/AdminManager.cs
@@ -279,7 +279,13 @@ namespace Content.Server.Administration.Managers
                 _commandPermissions.LoadPermissionsFromStream(efs);
             }
 
-            if (_res.TryContentFileRead(new ResPath("/toolshedEngineCommandPerms.yml"), out var toolshedPerms))
+            var toolshedPermsPath = new ResPath("/toolshedEngineCommandPerms.yml");
+            if (_res.TryContentFileRead(toolshedPermsPath, out var toolshedPerms))
+            {
+                _commandPermissions.LoadPermissionsFromStream(toolshedPerms);
+            }
+            // This may or may not be necessary. We read the same file again and load the same permissions into a different manager.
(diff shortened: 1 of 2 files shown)


DEATHB4DEFEAT:
> For reasons unknown

Is math done by the server?

Mnemotechnician:
> Is math done by the server?

Supposedly so, although we can easily lock these commands behind +ADMIN if needed.

DEATHB4DEFEAT:
> > Is math done by the server?
> 
> Supposedly so, although we can easily lock these commands behind +ADMIN if needed.

If the server is doing math it makes sense to restrict it.

Mnemotechnician:
The thing is, this is a bug that prevented anyone without +HOST from using any commands listed in that file. I can add an admin flag to these commands if you want specifically that.

Mnemotechnician:
Here.
image

DEATHB4DEFEAT (approved):

DangerRevolution (approved):

DangerRevolution (approved):

PR #574, opened 2024-07-21
People Can't Talk In Space

VMSolidus:
# Description

This fixes a bug that has been present in the game for ages, that people can talk to each other in the cold vacuum of space without the aid of a radio. I figured out that this doesn't actually need me to make my "SpaceAudioSystem" just yet, and can be handled with an extremely easy check of the voice source's Transform. Simply put, if they're not touching any grid, their voice can't be heard. Additionally, if a speaker is standing on a different grid from a potential receiver, their voice cannot be heard since the two are on different grids. Effectively, all verbal communication MUST be done through a radio if anyone involved is in space, or on a different grid. Voice can no longer travel between grids. 

Sound should not travel through the vacuum of space. Use a radio to communicate, or physically grab them, and hug them to within 3/10th of a tile. This extremely small distance allotment is there to simulate characters "Grabbing" each other in space to enable direct speaking communication.

This also means that in space, you can have a completely private conversation with someone by hugging them. Nukies would be able to discuss their attack plans while in the shadow of a station, without fear of being overheard by crew in defiance of really basic physics.

**I am coming for general audio next. We will restore the rule of physics and sanity to this game**

# Media

https://github.com/user-attachments/assets/16f107f5-86ab-480b-85df-92b79632e3c1

# Changelog

:cl:
- add: Voice will no longer travel through space. If you are in space, you must either have a radio to communicate with others, or get within hugging distance of someone to talk.

### DIFF
--- a/Content.Server/Chat/Systems/ChatSystem.cs
+++ b/Content.Server/Chat/Systems/ChatSystem.cs
@@ -505,7 +505,7 @@ public sealed partial class ChatSystem : SharedChatSystem
         var language = languageOverride ?? _language.GetLanguage(source);
         var languageObfuscatedMessage = SanitizeInGameICMessage(source, _language.ObfuscateSpeech(message, language), out var emoteStr, true, _configurationManager.GetCVar(CCVars.ChatPunctuation), (!CultureInfo.CurrentCulture.IsNeutralCulture && CultureInfo.CurrentCulture.Parent.Name == "en") || (CultureInfo.CurrentCulture.IsNeutralCulture && CultureInfo.CurrentCulture.Name == "en"));
 
-        foreach (var (session, data) in GetRecipients(source, Transform(source).GridUid == null ? 0.1f : WhisperMuffledRange))
+        foreach (var (session, data) in GetRecipients(source, Transform(source).GridUid == null ? 0.3f : WhisperMuffledRange))
         {
             if (session.AttachedEntity is not { Valid: true } listener)
                 continue;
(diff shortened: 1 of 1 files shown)


DEATHB4DEFEAT (approved):
Add media to the PR

stellar-novas:
So with this, people can still talk in space if they're on the same grid, ex. standing on solars?

FoxxoTrystan:
Two line changes, seriously?

Also is it not best to use the "close whisper range"?
(the range where you get the full whisper message)

VMSolidus:
> Add media to the PR

21mb over the limit, will post in discord

leonardo-dabepis:
Wouldn't sound go away if you're also just inside a spaced room? I know the building structure would pick up sound vibrations to some degree but I don't know if it'd be really audible. I have yet to try this out myself in real life.

VMSolidus:
> Two line changes, seriously?
> 
> Also is it not best to use the "close whisper range"? (the range where you get the full whisper message)

Surprisingly NOPE! Because the check is done at the obfuscate range, and not the close range!

DEATHB4DEFEAT:
> > Add media to the PR
> 
> 21mb over the limit, will post in discord

I've said several times to people, do not upload it to Discord, compress it with https://8mb.video and upload it to the PR.

VMSolidus:
> Wouldn't sound go away if you're also just inside a spaced room? I know the building structure would pick up sound vibrations to some degree but I don't know if it'd be really audible. I have yet to try this out myself in real life.

Sound transmits through the structure of the grid! But would be muffled, you're right. This however is a separate system I'll do, which involves making its own "Space Audio System". I'm doing that as a separate PR for all audio. It will involve math. Because naturally I know the math for exactly that. :)

FoxxoTrystan:
> > Two line changes, seriously?
> > Also is it not best to use the "close whisper range"? (the range where you get the full whisper message)
> 
> Surprisingly NOPE! Because the check is done at the obfuscate range, and not the close range!

Sooo, Normal Tallk and whisper does the same effect, and i supose sign language is bypassed as Emotes are not blocked that way.

VMSolidus:
> Add media to the PR

Done, thanks for the link. You've never sent that to me before, so I didn't know about it

DangerRevolution (approved):
Kinda crazy it's this shrimple to do

PR #580, opened 2024-07-22
Carrying System Refactor

VMSolidus:
# Description

This is a very simple and apparently minor update to the Carrying System, bringing it up to date with more modern code. The biggest difference is that rather than having a private one-off implementation of one of the original Nyano MassContest functions, it uses the new public Reworked MassContests. With this change, pick up durations no longer infinitely scale with arbitrary mass, meaning that a hypothetical 2000kg Lamia doesn't have an arbitrarily infinitesimal pickup duration when trying to pick up a 10kg Harpy. Carrying is also more strictly limited by mass, rather than by carrying duration, meaning that if a target character is more than 25% heavier than your character, it will not be possible to shoulder them. You'll just have to either drag them, or get a roller bed to move overly massive characters. 

The last thing I did was just cleanup all of the code, so that has nice, Single-IF exit conditions, rather than 30+ line blocks of IF(THING) RETURN;

Oh, and entities can now set their own internal base PickupDuration, so that entities can declare however easy or difficult they should be to pick up!

# MEDIA

https://github.com/user-attachments/assets/9ee0f1dd-ac75-406f-8bbd-9a130594d46d

# Changelog

:cl:
- tweak: The Carrying system has been reworked as a means of better supporting having extremely large species and characters. 10kg Harpies should no longer be oppressed by 2000kg Lamia with infinitely short carry attempts.


OldDanceJacket (approved):

Mnemotechnician:
How about instead of using a fixed maximum advantage of 1.25x, you make it an asymptote so that the advantage keeps growing along the mass ratio, but never exceeds something like 3x? Either that or it should be increased to at least 2x...

VMSolidus:
> How about instead of using a fixed maximum advantage of 1.25x, you make it an asymptote so that the advantage keeps growing along the mass ratio, but never exceeds something like 3x? Either that or it should be increased to at least 2x...

I'm actually currently working on something adjacent to this in a separate PR, my first ContestsSystem expansion. One of the things I was doing is adding an optional override that increases the function's Range as a multiplier. But that's actually a fantastic idea to make variant helpers that use more exotic curves.

VMSolidus:
> How about instead of using a fixed maximum advantage of 1.25x, you make it an asymptote so that the advantage keeps growing along the mass ratio, but never exceeds something like 3x? Either that or it should be increased to at least 2x...

In a separate PR that I'm not ready to post yet. :)

image

Mnemotechnician:
> I'm actually currently working on something adjacent to this in a separate PR, my first ContestsSystem expansion. One of the things I was doing is adding an optional override that increases the function's Range as a multiplier. But that's actually a fantastic idea to make variant helpers that use more exotic curves.

SS14 already has a tree of classes related to curves. They're used in HTN, but you can probably use them in your method.

VMSolidus:
> How about instead of using a fixed maximum advantage of 1.25x, you make it an asymptote so that the advantage keeps growing along the mass ratio, but never exceeds something like 3x? Either that or it should be increased to at least 2x...

image

It now scales with up to a 2x multiplier instead of a 1.25 multiplier.
However, it also simultaneously is impacted by StaminaContest now. If you are winded with stamina damage, it's harder. And if the victim has stamina damage, it gets easier.

image

Carry Slowdown is set to use the "Classic Mode" override, so that the carry slowdown has infinite scaling.

VMSolidus:
https://github.com/user-attachments/assets/25c12f05-7f1c-4fbf-b788-20020b069f19

I've just fixed two key bugs with the Carrying System. Now the ~1 year old bug where carried entities couldn't be thrown has been fixed, and once more you can throw felinids into trash cans. The other bug with players not having slowdowns applied correctly has also been fixed. You will now always be slowed down by carrying someone, even if you are heavier than them.

DangerRevolution (approved):

PR #589, opened 2024-07-24
Survival Is Unbearably Fast

VMSolidus:
# Description

I genuinely have no idea why we left the End Time at 40 minutes(Standard desired gamelength for NRP Servers), when the typical desired End Time for MRP is 2 hours. This coupled with an absurdly high chaos multiplier resulted in rounds where 40 minutes in, an event would fire every 30 seconds. 

I haven't touched any of the underlying math, because I can't be bothered writing a new equation for it today, but so help me god the Ramping Event Scheduler needs better math. And I'm considering that it should probably involve a Logistic Curve. 

# Changelog

:cl:
- tweak: Survival should no longer fire events every 30 seconds at the 40 minute mark. That may be normal on an NRP server, but we write with MRP+ in mind, where 120 minutes is typically the desired endpoint.

### DIFF
--- a/Content.Shared/CCVar/CCVars.cs
+++ b/Content.Shared/CCVar/CCVars.cs
@@ -100,7 +100,7 @@ namespace Content.Shared.CCVar
         /*
          * Events
          */
-
+        #region Event Timers
         /// <summary>
         ///     Controls if the game should run station events
         /// </summary>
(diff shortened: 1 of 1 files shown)


VMSolidus:
I'm closing this because I am sufficiently irritated by this system and its inflexibility that I am now getting out my graphing calculator.

DEATHB4DEFEAT:
> I genuinely have no idea why we left the End Time at 40 minutes(Standard desired gamelength for NRP Servers), when the typical desired End Time for MRP is 2 hours. This coupled with an absurdly high chaos multiplier resulted in rounds where 40 minutes in, an event would fire every 30 seconds.

https://github.com/Simple-Station/Einstein-Engines/pull/509

PR #593, opened 2024-07-25
Fix Talking Across Docked Grids

VMSolidus:
# Description

https://github.com/Simple-Station/Einstein-Engines/pull/574 introduced a bug whereby people who are each on separate grids, but the grids were attached, could not talk to each other. I have corrected this by making it so that when the Cross-Grid check is handled, it also checks to see if the two different grids are JOINTed to each other. Therefore allowing sound to travel across the connection. This should also work from Shuttle To Planet, since it's also handled via the same system. This means that docked shuttles allow sound to travel across to the docked station(or other shuttle), as well as shuttles docked to a planet's surface will permit sound to travel to the planet.

# Changelog

:cl:
- fix: Fixed a bug where sound was not traveling over a Shuttle-Docking connection. Attached grids now permit sound to travel through.

### DIFF
--- a/Content.Server/Chat/Systems/ChatSystem.cs
+++ b/Content.Server/Chat/Systems/ChatSystem.cs
@@ -748,7 +748,8 @@ public sealed partial class ChatSystem : SharedChatSystem
         var language = languageOverride ?? _language.GetLanguage(source);
         foreach (var (session, data) in GetRecipients(source, Transform(source).GridUid == null ? 0.3f : VoiceRange))
         {
-            if (session.AttachedEntity != null && Transform(session.AttachedEntity.Value).GridUid != Transform(source).GridUid
+            if (session.AttachedEntity != null
+                && Transform(session.AttachedEntity.Value).GridUid != Transform(source).GridUid
                 && !CheckAttachedGrids(source, session.AttachedEntity.Value))
                 continue;
 
@@ -986,10 +987,9 @@ public sealed partial class ChatSystem : SharedChatSystem
             return false;
 
         foreach (var (id, _) in sourceJoints.GetJoints)
-        {
             if (receiverJoints.GetJoints.ContainsKey(id))
                 return true;
-        }
+
         return false;
     }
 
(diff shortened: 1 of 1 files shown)


Content.Server/Chat/Systems/ChatSystem.cs
```diff
@@ -743,7 +748,8 @@ private void SendInVoiceRange(ChatChannel channel, string name, string message,
         var language = languageOverride ?? _language.GetLanguage(source);
         foreach (var (session, data) in GetRecipients(source, Transform(source).GridUid == null ? 0.3f : VoiceRange))
         {
-            if (session.AttachedEntity != null && Transform(session.AttachedEntity.Value).GridUid != Transform(source).GridUid)
+            if (session.AttachedEntity != null && Transform(session.AttachedEntity.Value).GridUid != Transform(source).GridUid
+                && !CheckAttachedGrids(source, session.AttachedEntity.Value))
```
DEATHB4DEFEAT, line 752:
```suggestion
            if (session.AttachedEntity != null
                && Transform(session.AttachedEntity.Value).GridUid != Transform(source).GridUid
                && !CheckAttachedGrids(source, session.AttachedEntity.Value))
```

Content.Server/Chat/Systems/ChatSystem.cs
```diff
@@ -973,6 +979,20 @@ public string BuildGibberishString(IReadOnlyList<char> charOptions, int length)
...
+    {
+        if (!TryComp<JointComponent>(Transform(source).GridUid, out var sourceJoints)
+            || !TryComp<JointComponent>(Transform(receiver).GridUid, out var receiverJoints))
+            return false;
+
+        foreach (var (id, _) in sourceJoints.GetJoints)
+        {
+            if (receiverJoints.GetJoints.ContainsKey(id))
+                return true;
+        }
```
DEATHB4DEFEAT, line 992:
```suggestion
        foreach (var (id, _) in sourceJoints.GetJoints)
            if (receiverJoints.GetJoints.ContainsKey(id))
                return true;

```

DEATHB4DEFEAT (approved):

OldDanceJacket (approved):

PR #605, opened 2024-07-28
NyanoCombat 2, Part 1: Simple Melee Refactor

VMSolidus:
# Description

Done in partnership with @OldDanceJacket 

This is a simple refactor of the Melee System, which introduces some new implementations of Contests, but more importantly lightly refactors melee so that "Light Attack"(Left Click) and "Heavy Attack"(Right Click) are now no longer hardcoded to be the same thing, and now can have different modifiers. We've set things up so that weapons can have different behaviors and bonuses if they are used in a heavy attack as opposed to a light attack, including range and damage. 

In addition to simple code cleanup, I have done the following things:
- Max Number of targets per Power Attack(Wide Swing) is now set via a CVar instead of a constant, but is still defaulted to 5.

All Blunt Weapons have been changed to
- Deal 100% of Blunt as Stamina(up from 50%) by default
- BluntDamageStaminaFactor can now be overridden by specific weapons

Light Attacks have been changed to:
- Deal up to 25% less damage scaling with taken stamina damage
- Deal up to 20% more damage with being injured

Heavy Attacks have been changed to:
- Deal up to 50% less damage scaling with taken stamina damage
- Deal up to 20% more damage with being injured
- Now deal (by default) 10 points of stamina damage to the wielder
- In a separate PR, nearly every weapon in the game is going to have their Wide Swing(Now "Power Attack") mode stats looked at.
- When examining the damage a weapon deals, if it has any damage modifiers when power attacking, these values will be shown in addition to the light attack damage.

Heavy attacks can now have separate modifiers for:
- Attack Range (multiply against light attack range)
- Attack Damage (multiply against light attack damage)
- Attack Speed (multiply against light attack speed)
- Stamina Cost (overriding the default)

We will be building upon this PR throughout the week with additional PRs, to include a full rework of melee weapon values across the entire board. For the most part, this PR is primarily here to give new options for contributors to balance melee weapons.

# Note For Maintainers:

@OldDanceJacket will be following this up with his own PR containing Part 2, which will include a comprehensive rework of the damage values for every melee weapon in the entire game. Due to this, I can't actually make that much in the way of significant changes to the components in review without messing with his side of things.

I also have my own Part 3 that I will be working in partnership with ODJ on throughout this week, which will consist of a pack of combat-system related Traits.

# Changelog

:cl: @VMSolidus and @OldDanceJacket 
- add: Being injured now lets you deal up to 20% more damage in melee combat to simulate the adrenaline rush from taking a hit.
- add: Taking stamina damage now makes you deal less damage in melee. Up to 25% less for Light Attacks, and 50% less for Power Attacks
- add: Wide Swing has been reworked as "Power Attack". Power attacks cost 20 stamina(by default) to perform, and their effects can differ from Light attacks on a per-weapon basis. Individual weapons can also have different stamina costs to power attack with.


DangerRevolution:
Shouldn't max targets per wide swing be a data field for weapons?

VMSolidus:
> Shouldn't max targets per wide swing be a data field for weapons?

it was originally a Constant, I think for either network performance or balance reasons, not sure, and it wasn't clear why it was a constant before. But just in case there was an inherent reason why it was a constant, to be safe I made it a CVar rather than a variable on weapons. But if that isn't the case, I can actually do both by making it the Min of component.MaxTargets and CCVars.MaxTargets.

DangerRevolution:
I'd honestly want to see Part 2 & Part 3 and especially hear about the philosophy behind these changes

VMSolidus:
> Shouldn't max targets per wide swing be a data field for weapons?

Done.

VMSolidus:
> > > Shouldn't max targets per wide swing be a data field for weapons?
> > 
> > 
> > Done.
> 
> Tests failing

Made a simple mistake, they should be passing now.

DangerRevolution:
Given this is a chonking PR, probably worth to have 3-4 approvals before merge :)

DangerRevolution (approved):

DangerRevolution:
Resolved conflict; stamina cost went from 10 --> 20

VMSolidus:
Lmao, this magically got turned into a single-line PR.

OldDanceJacket (approved):

PR #606, opened 2024-07-29
Radiation Doesn't Prevent DoAfters

VMSolidus:
# Description

I got so pissed off that being near a singularity means you can't crowbar doors, that shortly after dying to a singularity because of said bullshit, I had written this before an emergency shuttle even arrived to pick up the crew. 

# Changelog

:cl:
- fix: Radiation damage no longer interrupts DoAfters.

### DIFF
--- a/Content.Shared/DoAfter/SharedDoAfterSystem.cs
+++ b/Content.Shared/DoAfter/SharedDoAfterSystem.cs
@@ -65,7 +65,8 @@ public abstract partial class SharedDoAfterSystem : EntitySystem
     {
         // If we're applying state then let the server state handle the do_after prediction.
         // This is to avoid scenarios where a do_after is erroneously cancelled on the final tick.
-        if (!args.InterruptsDoAfters || !args.DamageIncreased || args.DamageDelta == null || GameTiming.ApplyingState)
+        if (!args.InterruptsDoAfters || !args.DamageIncreased || args.DamageDelta == null || GameTiming.ApplyingState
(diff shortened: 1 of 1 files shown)


DEATHB4DEFEAT (approved):

beck-thompson:
Shouldn't this also stop cold / heat damage? Seems like the same damage "type" if you know what I mean.

beck-thompson:
Mind telling me why this is a bad idea? It makes sense to me but two thumbs down tells me I must be missing something!

VMSolidus:
> Mind telling me why this is a bad idea? It makes sense to me but two thumbs down tells me I must be missing something!

So it's not necessary with cold because there's zero sources of cold damage in the game that deal enough damage in a single tick to interrupt do-afters. Fire damage I actually agree that it should because it's something "Sufficiently disruptive and avoidable" that getting caught by fire is genuinely your own fault. Radiation damage is completely invisible, with no feedback, and no indication at all of what's going on. Lord Singuloth can be completely offscreen, and still prevent you from crowbarring doors to get away from him. Supermatter can also get radioactive enough to interrupt do-afters from literally the opposite side of the station.

beck-thompson:
Mostly makes sense! Thanks for the response! I felt a little bad after the two thumbs down 😆

DangerRevolution (dismissed):
Changelog doesn't accurately represent PR

VMSolidus:
> Changelog doesn't accurately represent PR

You are a monster.

DangerRevolution (approved):
Not even evil, radiation just exists in more scenarios than Singulo

PR #630, opened 2024-08-01
Rename Fork Specific Options From "DeltaV" To "EinsteinEngines"

### LINKED ISSUE #481: Fork specific options are still called "Delta-V" in the settings
# Media

image

Tmanzxd:
Fixes #481


Content.Client/Options/UI/OptionsMenu.xaml.cs
```diff
@@ -20,7 +20,7 @@ public OptionsMenu()
             Tabs.SetTabTitle(2, Loc.GetString("ui-options-tab-controls"));
             Tabs.SetTabTitle(3, Loc.GetString("ui-options-tab-audio"));
             Tabs.SetTabTitle(4, Loc.GetString("ui-options-tab-network"));
-            Tabs.SetTabTitle(5, Loc.GetString("ui-options-tab-deltav")); // DeltaV specific settings
+            Tabs.SetTabTitle(5, Loc.GetString("ui-options-tab-einsteinengines")); // DeltaV specific settings
```
DEATHB4DEFEAT, line 23:
```suggestion
            Tabs.SetTabTitle(5, Loc.GetString("ui-options-tab-einsteinengines"));
```

DEATHB4DEFEAT (changes requested):
I think this should just be removed, it's only the species vision right? We have traits for that.

Content.Client/Options/UI/OptionsMenu.xaml.cs
Tmanzxd: Yea we could remove the tab outright if the species vision is not applied.

DangerRevolution:
> I think this should just be removed, it's only the species vision right? We have traits for that.

This goes back to the discussion about whether or not being able to cheaply opt out of species visions is okay or not. There's pros and cons to each side, but IMO it's for a different PR :)

DangerRevolution (changes requested):
Until consensus about removing the tab or not, revert the removal :)

Tmanzxd:
Got it

VMSolidus:
My concession is that while yes it's important to have Accessibility Options, we currently don't have ANY other than the vision toggle(And most of the really good accessibility features require Harmony....), so can we move the button to one of the pre-existing tabs? Also, we can set that CVar to default to false.

FoxxoTrystan:
Personnally il say move the option to another tab, simply.

DangerRevolution:
Test fail valid

Tmanzxd:
I can rebase the branch to the current master branch then do the tab rename if we want.

I can also remove the vision filter option as our traits cancel it out so the EE tab will be a placeholder for future additions.

DangerRevolution:
> I can rebase the branch to the current master branch then do the tab rename if we want.
> 
> I can also remove the vision filter option as our traits cancel it out so the EE tab will be a placeholder for future additions.

Leave the options, just rename the tab.

FoxxoTrystan:
Whats the update on that PR?

VMSolidus:
Closed since it's superceded by https://github.com/Simple-Station/Einstein-Engines/pull/928

PR #635, opened 2024-08-01
Make Boxes Not Useless

VMSolidus:
# Description

Subcontainers such as Boxes, Medkits, and Lunchboxes are essentially useless items that players always throw away. The given reason for this is that because the boxes only ever provided the same or less inventory slots than the inventory slots they take away, using boxes only ever penalizes players. Since you can only ever open one container at a time, placing items in boxes means you have to first open your backpack, then open the box, then take the item out.

How it SHOULD work instead, is that boxes contain slightly more inventory slots than what they remove. You would effectively be trading the ability to quickly and efficiently access all your items, for having slightly more inventory capacity. This PR buffs the parent of four different categories of items, such that 3x3 boxes occupy 9 inventory spaces and provide 16 spaces, and 2x4 boxes occupy 8 spaces and provide 12 spaces. The difference accounts for 3x3 boxes being significantly more inventory-awkward to put in your backpack than the 2x4 containers. 

Toolboxes are unchanged because they're already huge, and don't fit in backpacks anyway.

Here's a demonstration of inventory management with a backpack full of boxes.

https://github.com/user-attachments/assets/40b184f6-b061-4f0a-8e8b-7e2a7b9793bc

# Changelog

:cl:
- add: Boxes(Cardboard, Medkits, Lunchboxes) now contain slightly more inventory slots than they occupy.

### DIFF
--- a/Resources/Prototypes/Catalog/Fills/Boxes/general.yml
+++ b/Resources/Prototypes/Catalog/Fills/Boxes/general.yml
@@ -11,7 +11,7 @@
   - type: Storage
     maxItemSize: Small
     grid:
-    - 0,0,2,2
+    - 0,0,3,3
   - type: Sprite
     state: box
   - type: EmitSoundOnPickup
--- a/Resources/Prototypes/DeltaV/Entities/Objects/Consumable/Food/Containers/lunchbox.yml
+++ b/Resources/Prototypes/DeltaV/Entities/Objects/Consumable/Food/Containers/lunchbox.yml
@@ -18,9 +18,9 @@
   - type: Storage
     maxItemSize: Normal
     grid:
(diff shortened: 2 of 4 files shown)


DEATHB4DEFEAT:
Doesn't this trigger a test fail?

VMSolidus:
> Doesn't this trigger a test fail?

If it does I'm yeeting the test. But it shouldn't?

DEATHB4DEFEAT:
> > Doesn't this trigger a test fail?
> 
> If it does I'm yeeting the test. But it shouldn't?

Definitely used to.

OldDanceJacket (approved):

DEATHB4DEFEAT (approved):

PR #641, opened 2024-08-01
Make Secglasses Roundstart

angelofallars:
# Description

Cherry-picked from Space Wizards (https://github.com/space-wizards/space-station-14/pull/26487).

Original description:

> ## About the PR
> Replaces all instances of sunglasses & sechud in security lockers & job fills with secglasses.
> Removes the research for secglasses since it is now pointless. To make them obtainable past  roundstart, you can now craft secglasses by-hand using a sechud + glasses + some cable coil.
> ## Why / Balance
> This has been discussed a few times before in the past, and the conclusion I've come to is that the tradeoff of sechud / glasses isn't really a very meaningful or interesting one. Making the person using the HUD unable to use flashes or flashbangs doesn't really do anything other then giving them slightly less options.
> This is also pretty bad when you factor in that the sechud is pretty important QOL for a security officer. This ends up nerfing criminal records in a kind of unsavory way.

# Changelog

:cl: Flareguy
- tweak: Security glasses have been moved from research to roundstart gear. All officers now start with them instead of sunglasses by default.
- add: You can now craft security glasses.


DangerRevolution (changes requested):
Does this need an entry in migration files too maybe? To replace every SecHud with SecGlasses?

Also, this should be discussed by maints as this is a sussy change

DangerRevolution:
I'm going to close this by the weekend if no maintainer pitches in; as far as I can synthesise from discussions I've been in: this is has/is going to be put into loadouts as a half-compromise. Pay up points for Sec Glasses.

PR #645, opened 2024-08-02
Fix Event Manager Allowing to Start Multiple Events of the Same Type in a Short Time Window

Mnemotechnician:
# Description
Makes it so that the event scheduler doesn't pick events that were already added (but not yet ended).

In order to fix this:

<details><p>

image

</p></details>


Mnemotechnician:
This works, no more duplicate events:
image

DangerRevolution (approved):

DEATHB4DEFEAT (approved):
Shrimple as that

Issue #145, opened 2024-04-21
Failed to cherry-pick PR #25986: Return Detective to Security


SimpleStation14:
## Failed to cherry-pick PR: cannot create a tree from a not fully merged index.; class=Index (10); code=Unmerged (-10)
PR body below

###### `88d424e877de68e75f25c6efbed21aa92b264a59`

PR opened by <img src="https://avatars.githubusercontent.com/u/135308300?v=4" width="16"/><a href="https://github.com/LankLTE"> LankLTE</a> at 2024-03-11 03:14:52 UTC
PR merged by <img src="https://avatars.githubusercontent.com/u/19864447?v=4" width="16"/><a href="https://github.com/web-flow"> web-flow</a> at 2024-03-11 03:33:08 UTC

---

PR changed 9 files with 61 additions and 54 deletions.

The PR had the following labels:
- No C#
- Changes: Sprites

---

<details open="true"><summary><h1>Original Body</h1></summary>

> 
> 
> 
> ## About the PR
> 
> Reverts #23114 and #24739. Returns the detective to security, and any subsequent changes that come along with that (I believe those two PRs are the only ones to have changed anything, excluding maps which are fine to stay).
> 
> ## Why / Balance
> 
> It's run its' course. I've heard that detective usually ends up working for security anyways and as a role it's probably better fitting there. I also don't think independent detective has anything particularly interesting that warrants it staying.
> 
> - [X] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase
> 
> **Changelog**
> 
> 
> :cl: Lank
> - tweak: The detective is no longer independent, and again works under Security.
> 

</details>

Mnemotechnician:
This is fixed it seems? Should be closed now.

PR #656, opened 2024-08-03
Fixes Failed to Cherry-Pick PR #25986: Return Detective to Security #145

Tmanzxd:
Partial fix to #145, as the changes in inventory will need reflected in the Security Loadout system which will be in another PR.

Made so Detective has the same access as Security Officers.


DangerRevolution (approved):

VMSolidus (approved):
Utterly baffled why Security Officers have Cryogenics access? Unless that refers to Arrivals cryo. But oh well.

PR #679, opened 2024-08-06
Stamina Damage Resistance Real

whateverusername0:
# Description
Added stamina damage resistance, acts the same way as armor does.
Call this shock resistance if you wish :trollface:
Just attach StaminaDamageResistance component to an entity with a set multiplier and have fun.

Made all hardsuits 25% stun resistant by default.
With some variety, e.g. nukie, ERT, captain, HoS suits are 50%, DS are 90%, etc. etc.

This will not remove stuneta but it will make it more difficult to stamcrit a traitor or such.

Some armor/batong ratios that you need to hit before the target is stamcritted:
0% - 3 batong hits
25% - 4 batong hits
50% - 6 batong hits
75% - 12 batong hits
90% - 28 batong hits :trollface:
100% - 
image

# Changelog

:cl:
- add: Added different stamina damage resistance to hardsuits.

### DIFF
--- a/Content.Server/Nyanotrasen/Abilities/Boxer/BoxingSystem.cs
+++ b/Content.Server/Nyanotrasen/Abilities/Boxer/BoxingSystem.cs
@@ -14,7 +14,7 @@ public sealed partial class BoxingSystem : EntitySystem
         base.Initialize();
         SubscribeLocalEvent<BoxerComponent, ComponentInit>(OnInit);
         SubscribeLocalEvent<BoxerComponent, MeleeHitEvent>(OnMeleeHit);
-        SubscribeLocalEvent<BoxingGlovesComponent, StaminaMeleeHitEvent>(OnStamHit);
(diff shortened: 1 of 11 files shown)


VMSolidus:
image
I would like to let you know that this basically completely breaks those three systems by the way. They are *increasing outbound* stamina damage, and are dependent on having an event that links to *outbound* stamina damage.

VMSolidus:
Nevermind, carry on.
image

DangerRevolution:
Does it stack if you have multiple equipped?

whateverusername0:
> Does it stack if you have multiple equipped?

works just like normal armor but for stamina damage
so yes

DangerRevolution (approved):
would want to le playtest it but this is a code approval

VMSolidus (approved):
Been watching the playtests of it, this is pretty great! Thanks for also addressing the three instances we have of Nyano legacy code. I apologize you had to run into those. I still have my work cutout for me with refactoring the lot of them.

PR #686, opened 2024-08-06
New Trait: Blood Deficiency

angelofallars:
# Description

**Blood Deficiency** is a +2 points negative Physical trait that makes you slowly lose blood over time. When left untreated you will die from blood loss. Inspired by the SS13 trait of the same name.

Slash/Piercing weapons and bleeding are much more lethal against you.

The moment you start taking **any** blood loss damage, you start dying because you can't regenerate blood. Even just two consecutive kitchen knife stabs will make you bleed enough to die slowly unless you immediately apply gauze.

Blood packs, iron pills (or copper for Arachnids), and gauze to stop bleeding will help you survive with this trait.

Here's how the timeline looks for untreated blood deficiency:
- ~0-21 minutes: losing blood slowly
- ~21-31 minutes: blood level below 90%, start taking bloodloss damage
- ~31-33 minutes: critical
- ~34 minutes: death

## Media

<details><summary>Expand</summary>

**Trait entry**

image

image

As shown above, even just reducing the blood volume to less than 90% means you will die a slow and painful death.

</details>

</p>
</details>

# Changelog

:cl: Skubman
- add: Add the Blood Deficiency trait, a new negative trait that makes you slowly lose blood over time. You must routinely receive blood loss treatment to live, and even normally non-lethal bleeding can make you start dying slowly.

PR #689, opened 2024-08-07
Fix No Blood Regeneration

angelofallars:
# Description

Being a little too clever with omitting braces in complicated if statement chains in #686 caused blood regeneration for entities without Blood Deficiency to stop working altogether. I added some braces and now blood regeneration works again.

I also added `DataField` to `HasBloodDeficiency` because I forgot to add that.

# Changelog

:cl: Skubman
- fix: Passive blood regeneration now works again.


VMSolidus (approved):
Hilarious.

DangerRevolution (approved):

PR #692, opened 2024-08-07
New Trait: Sluggish / Snail-Paced

angelofallars:
# Description

This PR adds two new negative traits that decrease your movement speed: **Sluggish** and **Snail-Paced**.

- Sluggish (+1 points)
  - 15% slower movement speed
  - 35% slower climbing speed (for standard tables, from 1.5s to ~2.02 seconds)
  - Cooldown on laying down/standing up increased from 2.5 seconds to 3 seconds
- Snail-Paced (+2 points)
  - 30% slower movement speed
  - 66% slower climbing speed (for standard tables, from 1.5s to ~2.48 seconds)
  - Cooldown on laying down/standing up increased from 2.5 seconds to 4 seconds

## Media

<details><summary>Expand</summary>

**Trait entry**

image

image

</details>

# Changelog

:cl: Skubman
- add: Add two new negative traits: Sluggish (+1) and Snail-Paced (+2) that make you move slower, and climb tables slower.

### DIFF
--- a/Content.Server/Traits/Assorted/TraitSpeedModifierComponent.cs
+++ b/Content.Server/Traits/Assorted/TraitSpeedModifierComponent.cs
@@ -1,0 +1,14 @@
+namespace Content.Server.Traits.Assorted;
+
+/// <summary>
+///  This component is used for traits that modify movement speed.
+/// </summary>
+[RegisterComponent]
+public sealed partial class TraitSpeedModifierComponent : Component
+{
+    [DataField("walkModifier", required: true)]
+    public float WalkModifier = 1.0f;
+
+    [DataField("sprintModifier", required: true)]
+    public float SprintModifier = 1.0f;
+}
--- a/Content.Server/Traits/Assorted/TraitSpeedModifierComponent.cs
+++ b/Content.Server/Traits/Assorted/TraitSpeedModifierComponent.cs
@@ -6,7 +6,7 @@ namespace Content.Server.Traits.Assorted;
 [RegisterComponent]
 public sealed partial class TraitSpeedModifierComponent : Component
 {
-    [DataField("walkModifier", required: true)]
+    [DataField(required: true)]
     public float WalkModifier = 1.0f;
 
     [DataField("sprintModifier", required: true)]
--- a/Content.Server/Traits/Assorted/TraitSpeedModifierComponent.cs
+++ b/Content.Server/Traits/Assorted/TraitSpeedModifierComponent.cs
@@ -9,6 +9,6 @@ public sealed partial class TraitSpeedModifierComponent : Component
     [DataField(required: true)]
     public float WalkModifier = 1.0f;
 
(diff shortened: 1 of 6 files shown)


DangerRevolution (approved):
Dionabros we're so back!

DEATHB4DEFEAT:
> Dionabros we're so back!

Dionae are show on newPark (and old), kinda wanna port it here.

VMSolidus:
@angelofallars I have a fun question. A long while ago, I was considering doing a "Small rework" of Diona on DeltaV, and my planned idea was to make it so that Diona have a "Slow & Purposeful" species trait. Slow & Purposeful makes them always 25% slower to move, and can never slip, but with a tradeoff that they are completely immune to all movement modifiers. A Diona would be 25% slower wearing a jugsuit, wearing a duffel bag, holding one duffel bag in each arm, and walking through a puddle of glue covered in spider webs. 

Do you think this is something you would be able to do? Also for obvious reasons, Diona wouldn't be able to take other traits that modify movement.

angelofallars:
> @angelofallars I have a fun question. A long while ago, I was considering doing a "Small rework" of Diona on DeltaV, and my planned idea was to make it so that Diona have a "Slow & Purposeful" species trait. Slow & Purposeful makes them always 25% slower to move, and can never slip, but with a tradeoff that they are completely immune to all movement modifiers. A Diona would be 25% slower wearing a jugsuit, wearing a duffel bag, holding one duffel bag in each arm, and walking through a puddle of glue covered in spider webs.
> 
> Do you think this is something you would be able to do? Also for obvious reasons, Diona wouldn't be able to take other traits that modify movement.

That'd be a un trait to do, and an interesting take on the "Dionae are slow" thing. I'm of course down for doing that!

Content.Server/Traits/Assorted/TraitSpeedModifierComponent.cs
```diff
@@ -0,0 +1,14 @@
+namespace Content.Server.Traits.Assorted;
+
+/// <summary>
+///  This component is used for traits that modify movement speed.
+/// </summary>
+[RegisterComponent]
+public sealed partial class TraitSpeedModifierComponent : Component
+{
+    [DataField("walkModifier", required: true)]
```
VMSolidus, line 9:
```suggestion
    [DataField(required: true)]
```

Content.Server/Traits/Assorted/TraitSpeedModifierComponent.cs
```diff
@@ -0,0 +1,14 @@
...
+    public float WalkModifier = 1.0f;
+
+    [DataField("sprintModifier", required: true)]
```
VMSolidus, line 12:
```suggestion
    [DataField(required: true)]
```

VMSolidus (changes requested):

VMSolidus (approved):

PR #693, opened 2024-08-07
Melee Pt2

OldDanceJacket:
# PT2 of Melee Weapons The Numbers Don't Lie

This is part 2 of the ongoing work of Solid and myself going through and touching up the melee combat in the game. In this part I rebalance all of the melee weapons to generally do less damage, more stamina damage, and be more unique in regards to slight range changes, attack speed adjustments, along with every weapon getting slightly adjusted heavy swing changes ranging from attack rates, damage, range, angle, and how many targets you can hit.

Majority of weapons will hit the standard amount of targets of 5(the old norm), but a few are lowered to be single target hits. These are usually tightened in the angle that they attack in(old angle range was 60). Similarly all melee weapons have individual stamina costs on their heavy swings, most of these are in the range of 5 or 10, and following this PR the new standard should be 10 as the outliers that would abuse this have been addressed in this PR.

---

# Changelog

Normally I would do a changelog but this took awhile and I forgo. 

:cl: ODJ
- tweak: Melee Weapons now feel different across the board, from the Wrench to the Chainsaw, try out their normal swings and their heavy attacks!

### DIFF
--- a/Content.Client/Weapons/Melee/MeleeWeaponSystem.cs
+++ b/Content.Client/Weapons/Melee/MeleeWeaponSystem.cs
@@ -1,5 +1,6 @@
 using System.Linq;
 using Content.Client.Gameplay;
+using Content.Shared.CCVar;
 using Content.Shared.CombatMode;
 using Content.Shared.Effects;
 using Content.Shared.Hands.Components;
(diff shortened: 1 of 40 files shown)


gluesniffler:
wizmoids fear the spear
image

VMSolidus:
image
Test fails are legitimate, it looks like you accidentally messed up the yml entry for Pickaxe.

VMSolidus (approved):

DangerRevolution (approved):
Approved but I want to merge 1-3 at the same time.

PR #747, opened 2024-08-18
Add LocalizedPrototype Type

DEATHB4DEFEAT:
# Description

Adds a new Type for Prototypes to inherit instead of IPrototype with standard localization formats and lambda shortcuts to them.

---

### DIFF
--- a/Content.Shared/Prototypes/LocalizedPrototype.cs
+++ b/Content.Shared/Prototypes/LocalizedPrototype.cs
@@ -10,8 +10,10 @@ public abstract class LocalizedPrototype : IPrototype
 
     public const string LocFormat = "{type}-{ID}-{field}";
 
-    public string Name => ToLocalizationString("name");
-    public string LocalizedName => Loc.GetString(Name);
+    /// <summary>The localization string for the name of this prototype</summary>
+    public string NameLoc => ToLocalizationString("name");
+    /// <summary>The localized string for the name of prototype</summary>
+    public string Name => Loc.GetString(NameLoc);
 
     public string ToLocalizationString(string field)
     {
--- a/Content.Shared/Prototypes/LocalizedPrototype.cs
+++ b/Content.Shared/Prototypes/LocalizedPrototype.cs
@@ -8,13 +8,17 @@ public abstract class LocalizedPrototype : IPrototype
     [IdDataField]
     public string ID { get; } = default!;
 
-    public const string LocFormat = "{type}-{ID}-{field}";
+    public const string LocFormat = "{0}-{1}-{2}";
 
     /// <summary>The localization string for the name of this prototype</summary>
     public string NameLoc => ToLocalizationString("name");
     /// <summary>The localized string for the name of prototype</summary>
     public string Name => Loc.GetString(NameLoc);
 
+    /// <summary>
+    ///     Returns an Loc string using the <see cref="field"/> as the 'property'.
+    ///     Given `desc` it will return `this-prototype-PrototypeId-desc`.
+    /// </summary>
     public string ToLocalizationString(string field)
     {
         // Get the ID of the proto Type
@@ -26,12 +30,6 @@ public abstract class LocalizedPrototype : IPrototype
         // Replace every uppercase letter with a dash and the lowercase letter
         type = type.Aggregate("", (current, c) => current + (char.IsUpper(c) ? "-" + char.ToLowerInvariant(c) : c.ToString()));
 
-        // Replace the placeholders with the actual values
-        var t = LocFormat
-            .Replace("{type}", type)
-            .Replace("{ID}", ID)
-            .Replace("{field}", field);
-
-        return t;
+        return string.Format(LocFormat, type, ID, field);
     }
 }
--- a/Content.Shared/Prototypes/LocalizedPrototype.cs
+++ b/Content.Shared/Prototypes/LocalizedPrototype.cs
@@ -32,8 +32,6 @@ public abstract class LocalizedPrototype : IPrototype
             .Replace("{ID}", ID)
             .Replace("{field}", field);
 
-        Logger.Fatal(t);
-        Logger.Fatal(Loc.GetString(t));
         return t;
     }
 }
--- a/Content.Shared/Prototypes/LocalizedPrototype.cs
+++ b/Content.Shared/Prototypes/LocalizedPrototype.cs
@@ -1,0 +1,37 @@
+using System.Linq;
+using Robust.Shared.Prototypes;
+
+namespace Content.Shared.Prototypes;
+
+public abstract class LocalizedPrototype : IPrototype
+{
+    [IdDataField]
+    public string ID { get; } = default!;
+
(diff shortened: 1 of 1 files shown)


Content.Shared/Prototypes/LocalizedPrototype.cs
```diff
@@ -0,0 +1,37 @@
...
+    public string ID { get; } = default!;
+
+    public const string LocFormat = "{type}-{ID}-{field}";
+
+    /// <summary>The localization string for the name of this prototype</summary>
+    public string NameLoc => ToLocalizationString("name");
+    /// <summary>The localized string for the name of prototype</summary>
+    public string Name => Loc.GetString(NameLoc);
+
+    public string ToLocalizationString(string field)
```
Pspritechologist, line 18:
```suggestion
    /// <summary>
    ///     Returns an Loc string using the input as the 'property.
    ///     Given `desc` it will return `ThisPrototype-PrototypeId-desc`.
    ///     These strings can be used to automatically localize Prototypes.
    /// </summary>
    public string PropertyLocString(string field)
```

Content.Shared/Prototypes/LocalizedPrototype.cs
```diff
@@ -0,0 +1,37 @@
...
+{
+    [IdDataField]
+    public string ID { get; } = default!;
+
+    public const string LocFormat = "{type}-{ID}-{field}";
+
+    /// <summary>The localization string for the name of this prototype</summary>
+    public string NameLoc => ToLocalizationString("name");
+    /// <summary>The localized string for the name of prototype</summary>
+    public string Name => Loc.GetString(NameLoc);
```
Pspritechologist, line 16:
```suggestion
    public string Name => Loc.TryGetString(NameLoc) ?? Name;
    
    [DataField]
    private string _name = "";
```

Content.Shared/Prototypes/LocalizedPrototype.cs
```diff
@@ -0,0 +1,37 @@
...
+    public string ToLocalizationString(string field)
+    {
+        // Get the ID of the proto Type
+        var type =
+            ((PrototypeAttribute?) Attribute.GetCustomAttribute(GetType(), typeof(PrototypeAttribute)))?.Type
+            ?? GetType().Name.Remove(GetType().Name.Length - 9);
+        // Lowercase the first letter
+        type = char.ToLowerInvariant(type[0]) + type[1..];
+        // Replace every uppercase letter with a dash and the lowercase letter
+        type = type.Aggregate("", (current, c) => current + (char.IsUpper(c) ? "-" + char.ToLowerInvariant(c) : c.ToString()));
```
Pspritechologist, line 27: Document like, each microgram of this code. It's really gross and no one should need to read it.

Content.Shared/Prototypes/LocalizedPrototype.cs
```diff
@@ -0,0 +1,37 @@
...
+using Robust.Shared.Prototypes;
+
+namespace Content.Shared.Prototypes;
+
+public abstract class LocalizedPrototype : IPrototype
+{
+    [IdDataField]
+    public string ID { get; } = default!;
+
+    public const string LocFormat = "{type}-{ID}-{field}";
```
Pspritechologist, line 11:
```suggestion
    public const string LocFormat = "{}-{}-{}";
```

Content.Shared/Prototypes/LocalizedPrototype.cs
```diff
@@ -0,0 +1,37 @@
...
+        // Replace every uppercase letter with a dash and the lowercase letter
+        type = type.Aggregate("", (current, c) => current + (char.IsUpper(c) ? "-" + char.ToLowerInvariant(c) : c.ToString()));
+
+        // Replace the placeholders with the actual values
+        var t = LocFormat
+            .Replace("{type}", type)
+            .Replace("{ID}", ID)
+            .Replace("{field}", field);
+
+        return t;
```
Pspritechologist, line 35:
```suggestion
        return String.Format(LocFormat, type, ID, field);
```

Pspritechologist (dismissed):

Content.Shared/Prototypes/LocalizedPrototype.cs
DEATHB4DEFEAT: I do not like this.

VMSolidus (approved):

Content.Shared/Prototypes/LocalizedPrototype.cs
Pspritechologist: This opens up major flexibility for downstreams who may not want to do everything 'properly'.
It's quite a lot more effort to make individual FTL files for every single Prototype made and is only useful at large scale when planning for people who may fork and translate. Even then it still doesn't benefit *you*, only your potential downstreams. It may not matter for EE if the plan is to *require* the use of FTL files but that's still something we can just enforce while ensuring our systems are as user friendly as possible to people who don't care.

FoxxoTrystan (approved):

Content.Shared/Prototypes/LocalizedPrototype.cs
```diff
@@ -0,0 +1,35 @@
...
+{
+    [IdDataField]
+    public string ID { get; } = default!;
+
+    public const string LocFormat = "{0}-{1}-{2}";
+
+    /// <summary>The localization string for the name of this prototype</summary>
+    public string NameLoc => ToLocalizationString("name");
+    /// <summary>The localized string for the name of prototype</summary>
+    public string Name => Loc.GetString(NameLoc);
```
Mnemotechnician, line 16: This recalculates NameLoc every time either of those properties is accessed, which is bad. It should be cached and only calculated once (perhaps as a deserialization hook, or a lazily evaluated/cached property?)

Content.Shared/Prototypes/LocalizedPrototype.cs
```diff
@@ -0,0 +1,35 @@
...
+    public string ToLocalizationString(string field)
+    {
+        // Get the ID of the proto Type
+        var type =
+            ((PrototypeAttribute?) Attribute.GetCustomAttribute(GetType(), typeof(PrototypeAttribute)))?.Type
+            ?? GetType().Name.Remove(GetType().Name.Length - 9);
+        // Lowercase the first letter
+        type = char.ToLowerInvariant(type[0]) + type[1..];
+        // Replace every uppercase letter with a dash and the lowercase letter
+        type = type.Aggregate("", (current, c) => current + (char.IsUpper(c) ? "-" + char.ToLowerInvariant(c) : c.ToString()));
```
Mnemotechnician, line 31: This performs `n` string concatenations, each concatenation having a cost of `n` itself, resulting in an `O(n²)` operation. This should use a StringBuilder instead. Considering that this function will be called hundreds or thousands of times, performance matters here.

Mnemotechnician (changes requested):

FoxxoTrystan:
This PR is approved but is it still going forward?

PR #757, opened 2024-08-20
Implement Mind Contests

VMSolidus:
# Description

Now that the initial Psionic Refactor is out of the way, this is my first new standalone Psionics PR, filling in the MindContest function that was added over a month ago(but was set to always return 1f as a placeholder). 

Mind Contests are unique among the Contests System in that they do not treat the lack of a Psionic Component as a failure condition, and instead as a variable.

No changelog because MindContests are not currently used anywhere, all I'm doing is adding the actual function logic.

### DIFF
--- a/Content.Shared/Contests/ContestsSystem.cs
+++ b/Content.Shared/Contests/ContestsSystem.cs
@@ -30,7 +30,6 @@ namespace Content.Shared.Contests
         /// <summary>
         ///     Outputs the ratio of mass between a performer and the average human mass
         /// </summary>
-        /// <param name="performerUid">Uid of Performer</param>
         public float MassContest(EntityUid performerUid, bool bypassClamp = false, float rangeFactor = 1f, float otherMass = AverageMass)
         {
             if (!_cfg.GetCVar(CCVars.DoContestsSystem)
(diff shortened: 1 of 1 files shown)


Content.Shared/Contests/ContestsSystem.cs
```diff
@@ -41,7 +48,8 @@ public float MassContest(EntityUid performerUid, bool bypassClamp = false, float
 
         /// <inheritdoc cref="MassContest(EntityUid, bool, float, float)"/>
         /// <remarks>
-        ///     MaybeMassContest, in case your entity doesn't exist
+        ///     MaybeMassContest, in case your entity doesn'so be ready to meet god if you bypass the clamp.
+        /// </remarks> exist
```
DEATHB4DEFEAT, line 52:
```suggestion
```

Content.Shared/Contests/ContestsSystem.cs
```diff
@@ -41,7 +48,8 @@ public float MassContest(EntityUid performerUid, bool bypassClamp = false, float
 
         /// <inheritdoc cref="MassContest(EntityUid, bool, float, float)"/>
         /// <remarks>
-        ///     MaybeMassContest, in case your entity doesn't exist
+        ///     MaybeMassContest, in case your entity doesn'so be ready to meet god if you bypass the clamp.
```
DEATHB4DEFEAT, line 51: ?
```suggestion
        ///     MaybeMassContest, in case your entity isn't ready to meet god if you bypass the clamp.
```

Content.Shared/Contests/ContestsSystem.cs
DEATHB4DEFEAT, line 1: Remove all the empty param tags

DEATHB4DEFEAT (changes requested):

DEATHB4DEFEAT (approved):

OldDanceJacket (approved):

PR #763, opened 2024-08-21
Psionic Powers Add Components With Arguments.

VMSolidus:
# Description

While coding another Psionic Power feature, I discovered that the current implementation of iterating over components does not carry over arguments for the components. So I copied over the method used by Traits exactly-as-is, and just changed the names of the variables to accommodate the PsionicSystem.

# Changelog

:cl:
- add: PsionicPowers that add a Component now also allow for adding a Component with Arguments. This works exactly like the trait system's implementation of components.


DEATHB4DEFEAT (approved):

DangerRevolution (approved):
Probably doesn't need a CL but I'm still merging

PR #765, opened 2024-08-21
Traits Refactor

VMSolidus:
# Description

I decided traits weren't flexible enough, so I refactored them to also optionally add Actions(Activatable Abilities), and PsionicPowers(Going through the PsionicAbilitiesSystem). Neither of these have any current implementations, I'll leave that to other people. Trait Components are by extension no longer a hard requirement, although if you add a blank trait that adds nothing, you have only yourself to blame. But doing so won't crash the game or throw an error anyway.

# Changelog

:cl:
- add: Traits can now add Active Abilities to a character.
- add: Traits can now add Psionic Powers to a character.

### DIFF
--- a/Content.Server/Traits/TraitSystem.cs
+++ b/Content.Server/Traits/TraitSystem.cs
@@ -49,10 +49,20 @@ public sealed class TraitSystem : EntitySystem
                 out _))
                 continue;
 
-            AddTraitComponents(args.Mob, traitPrototype);
-            AddTraitActions(args.Mob, traitPrototype);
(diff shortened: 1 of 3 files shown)


Content.Server/Traits/TraitSystem.cs
DEATHB4DEFEAT, line 1: These should be public functions with one that calls them all

DangerRevolution (approved):

DEATHB4DEFEAT (changes requested):

DEATHB4DEFEAT (approved):

PR #775, opened 2024-08-21
Increase Trait Point Granularity

VMSolidus:
# Description

Since we now have a LOT of Traits, and soon to be another fresh batch of Traits coming in now that they've been refactored to offer even more options, it's come up that Traits are kinda hard to balance with the current combination of point costs and allotted starting points. To help alleviate this problem, I've both doubled the point values of every trait in the game, as well as doubled the default allotted starting points. There is now more room to balance trait costs against each other.

I have not however actually made use of the new range of point costs yet, and would like to consult with other contributors and maintainers on which traits need to be adjusted. 

# TODO

- [ ] Go over trait point costs again to address balance.
- [ ] Do this again in 6 months when we have 200 traits.

# Changelog

:cl:
- tweak: Trait points have been made more granular by both doubling the available number of trait points, and increasing the base cost of all pre-existing traits.

### DIFF
--- a/Content.Shared/CCVar/CCVars.cs
+++ b/Content.Shared/CCVar/CCVars.cs
@@ -367,7 +367,7 @@ namespace Content.Shared.CCVar
         ///     How many points a character should start with.
         /// </summary>
         public static readonly CVarDef<int> GameTraitsDefaultPoints =
-            CVarDef.Create("game.traits_default_points", 5, CVar.REPLICATED);
+            CVarDef.Create("game.traits_default_points", 10, CVar.REPLICATED);
 
 
         /// <summary>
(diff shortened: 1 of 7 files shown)


DEATHB4DEFEAT (approved):

DangerRevolution:
why does this need discussion if the PR makes no effective changes?

DangerRevolution (approved):

VMSolidus:
> why does this need discussion if the PR makes no effective changes?

I mean merge it if you want, that just means we'll need to update the points of traits to take advantage of this PR, in its own separate PR.

OldDanceJacket:
Bigger points pool is easier to balance due to having more to work with, but even though you're joking about doing this again in the future--It should never be raised again.

PR #774, opened 2024-08-21
Psi-Potentiometry Power

VMSolidus:
# Description

Psi-Potentiometry is a power pretty similar to PsionicRefactor-V1's Narrow Metapsionic Pulse, and is dual-typed as both a Psionic Power, as well as a system that can be used via an item of power. Although such an item does not currently exist as of this draft PR.
Potentiometry is intended to be a long-Doafter scan, which provides a UI window containing a character's list of MetapsionicFeedback strings.

# TODO

- [ ] Complete the code
- [ ] Maybe ask Leo to make a Psi-Potentiometer sprite, and then code it.

# Changelog

:cl:
- add: Potentiometry has been added as a new Psionic Power.

PR #776, opened 2024-08-21
Psionic Power Feedback Messages

VMSolidus:
# Description

This PR implements the previously planned feature whereby obtaining a Psionic Power plays some form of notification to alert the player that they have gained a new ability. Since some Psionics like Xenoglossy are purely passive, it's very important to give an indication to players what's going on. To that end, PsionicPowerPrototype has been expanded to include new datafields related to Initialization Feedback. There are now three kinds of feedback messages: Popup, Feedback, and Metapsionic.

All feedback will only play for powers obtained during the round, rather than for entities that innately start with powers.

- Popups will appear over your character's head as a small, brief message. These should be no more than a sentence at most.
- Feedback will appear in the Chat window as a message only visible to the Psion themself. These can be as much as a paragraph in length.
- Metapsionic messages are coming in their own separate PR: https://github.com/Simple-Station/Einstein-Engines/pull/774

In a separate PR, I also wish to add Audio feedback as well.

# Media

I apologize that the video has been bitcrunched to a point that the chat window can't be read.

https://github.com/user-attachments/assets/11e30e91-8fc6-48a2-b6a5-9ecf7127065e

# Changelog

:cl:
- add: Gaining a new Psionic Power can now display messages to alert the player, both as a short popup, and optionally a lengthier message sent to the user's Chat window.

### DIFF
--- a/Content.Server/Abilities/Psionics/PsionicAbilitiesSystem.cs
+++ b/Content.Server/Abilities/Psionics/PsionicAbilitiesSystem.cs
@@ -285,7 +285,7 @@ namespace Content.Server.Abilities.Psionics
             if (proto.InitializationFeedback is null)
                 return;
 
-            var feedbackMessage = Loc.GetString($"[font size={proto.InitializationFeedbackFontSize}][color={proto.InitializationFeedbackColor}]{proto.InitializationFeedback}[/color][/font]");
+            var feedbackMessage = $"[font size={proto.InitializationFeedbackFontSize}][color={proto.InitializationFeedbackColor}]{Loc.GetString(proto.InitializationFeedback)}[/color][/font]";
             _chatManager.ChatMessageToOne(
                 proto.InitializationFeedbackChannel,
                 feedbackMessage,
--- a/Content.Server/Abilities/Psionics/PsionicAbilitiesSystem.cs
+++ b/Content.Server/Abilities/Psionics/PsionicAbilitiesSystem.cs
@@ -285,7 +285,9 @@ namespace Content.Server.Abilities.Psionics
             if (proto.InitializationFeedback is null)
                 return;
 
-            var feedbackMessage = $"[font size={proto.InitializationFeedbackFontSize}][color={proto.InitializationFeedbackColor}]{Loc.GetString(proto.InitializationFeedback)}[/color][/font]";
+            if (!Loc.TryGetString(proto.InitializationFeedback, out var feedback))
+                return;
+            var feedbackMessage = $"[font size={proto.InitializationFeedbackFontSize}][color={proto.InitializationFeedbackColor}]{feedback}[/color][/font]";
             _chatManager.ChatMessageToOne(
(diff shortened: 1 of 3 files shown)


DangerRevolution (approved):
What's the use case for displaying a message in chat but not over your head, excluding situations where you use both

Content.Server/Abilities/Psionics/PsionicAbilitiesSystem.cs
```diff
@@ -290,11 +266,38 @@ private void AddPsionicStatSources(PsionicPowerPrototype proto, PsionicComponent
...
+                return;
+
+            if (proto.InitializationPopup is null)
+                _popups.PopupEntity(Loc.GetString(GenericInitializationMessage), uid, uid, PopupType.MediumCaution);
+            else _popups.PopupEntity(Loc.GetString(proto.InitializationPopup), uid, uid, PopupType.MediumCaution);
+
+            if (proto.InitializationFeedback is null)
+                return;
+
+            var feedbackMessage = $"[font size={proto.InitializationFeedbackFontSize}][color={proto.InitializationFeedbackColor}]{Loc.GetString(proto.InitializationFeedback)}[/color][/font]";
```
Mnemotechnician, line 288: It's probably better to do
```suggestion
            if (!Loc.TryGetString(proto.InitializationFeedback, out var feedback))
                return;
            var feedbackMessage = $"[font size={proto.InitializationFeedbackFontSize}][color={proto.InitializationFeedbackColor}]{feedback}[/color][/font]";
```
so you don't display invalid feedbacks

VMSolidus: I genuinely didn't know this was an option. NEAT!

Content.Shared/Psionics/PsionicPowerPrototype.cs
```diff
@@ -36,10 +37,37 @@ public sealed partial class PsionicPowerPrototype : IPrototype
...
+    ///     These should generally also be written in the first person, and can be far lengthier than popups.
+    /// </summary>
+    [DataField]
+    public string? InitializationFeedback;
+
+    /// <summary>
+    ///     What color will the initialization feedback display in the chat window with.
+    /// </summary>
+    [DataField]
+    public string InitializationFeedbackColor = "#8A00C2";
```
DEATHB4DEFEAT, line 58: Make this something that contrasts better with chat

DEATHB4DEFEAT (approved):

Content.Shared/Psionics/PsionicPowerPrototype.cs
VMSolidus: image

I'll take it one more step up the chain. That should be good. :)

ShatteredSwords:
> Oh damnit I hit the wrong button.

Who could have forseen these unfortunate circumstances?

PR #787, opened 2024-08-22
Delete All Then Ghost Fault Tolerance

VMSolidus:
# Description

image

I'm putting this here as an option to deal with our Heisentest problems, by making the tests "Fault-Tolerant" wherever practical, but I don't want this merged without Death and Psprite agreeing to this. For the most part I believe that these tests are failing because they are essentially checking that "Random events are not creating entities", by creating their own enforced Race Conditions. This particular test is repeatedly failing because the Mood System has no way of deducing that it's in a test. Even though the alleged issue is a nothingburger. 

image

Tests absolutely shouldn't have been designed around race conditions.

# Changelog

No changelog because this isn't playerfacing.


FoxxoTrystan (approved):
Pray

PR #801, opened 2024-08-25
Roboticist

Mnemotechnician:
# Description
By a (somewhat) popular request from floofstation, this adds the roboticist job (each existing map gets 1 job slot, since robotics on each of them is just one tiny room that's more similar to a broom closet than a subdepartament).

The job is purely cosmetical and does not offer any new accesses since that would require changing a lot of maps and would probably make being a scientist in the absense of a roboticist miserable. If needed, roboticists can always build blast doors on the entrance to robotics, as someone has done before in my experience.

This also adds the robotics PDA, robotics ID card, robotics satchel, backpack, and duffel (duffel icon taken from paradise station, rest of the sprites is respective science bags recolored by me to match the palette of the icon). This also increases the capacity of mystagogue's science PDA box to incorporate the new PDA type.

<p>

image

image

image

image

image

</p>
</details>

---

# Changelog
:cl:
- add: A new Roboticist job has been added. The job does not offer any new accesses, but lets you distinguish yourself from scientists. Existing maps will need to be updated to support the new job.
- add: Added new PDA, ID, and bag variants for roboticists. The mystagogue will now also find a roboticist PDA in their office.


DangerRevolution (changes requested):
make a spawner for robotocist

DangerRevolution (approved):

Resources/Prototypes/Maps/arena.yml
```diff
@@ -65,6 +65,7 @@
             ResearchAssistant: [ 2, 4 ]
             Chaplain: [ 1, 1 ]
             Borg: [ 2, 2 ]
+            Roboticist: [ 1, 1 ]
```
VMSolidus, line 68:
```suggestion
```
Remove the Roboticist entry from all maps, these would need to be in individual PRs for maps, since adding a job spawner requires that you touch a map to add it somewhere.

VMSolidus (changes requested):
REmove

Mnemotechnician:
ok

VMSolidus (approved):

Elijahrane:
Should be part of engineering

OldDanceJacket:
> Should be part of engineering

If we're having golems and robots, then you should move robots to engineering, Epi is really big, I agree with this actually.

Mnemotechnician:
> Should be part of engineering

Absolutely no.

DEATHB4DEFEAT:
Should be part of engineering

FoxxoTrystan:
We need normal Science back and just EPI be like Robotist, a side of Science and not a compleat replacement.

OldDanceJacket (approved):

OldDanceJacket:
Not going to let this sit because barely any discussion was had on the topic, it has no mechanics at all to it, so it can be merged instead of pointlessly sitting here dead.



> We need normal Science back and just EPI be like Robotist, a side of Science and not a compleat replacement.

The current direction of EE is psionics, so Epi has replaced Science and anything in the future should take that in mind. This is the direction we're currently going in, if you feel like this shouldn't be the case then make a discussion about it on discord. As it sits one of the big contributors(Solid) has been making Psionic focused PRs, and a number of PRs from other users have been returning missing content from Nyano that was lost in the DeltaV rebase.

PR #809, opened 2024-08-28
Make Shoving and Stamina Great Again

Mnemotechnician:
# Description
First off, shoving was broken badly. It would roll shoving chance twice, meaning that you had only about 6% chance to actually shove someone, and it would only remove 6% of their stamina. Additionally, stamcrits have always pissed me off by just how stupid they were. Also, some of the code in the stamina system had highly misleading names, which has led vmsolidus to implement mass contests the wrong way there.

This PR introduces changes to fix those issues:
- Shoving only rolls the shove chance once.
- Shoving deals 50 * shove chance stamina damage, as intended, without depending on target's stamcrit threshold. In the future it should depend on the weapon used (claws, fists, etc), but for now it's fine.
- Shoving advantage ranges were re-evaluated. Mass difference now can give 0.5x-2x advantage, and health difference can give 0.75x - 1.25x. Stamina difference still gives a neglectable 0.9x - 1.1x advantage.
- The stamina slowdown is now added and calculated dynamically using MovementSpeedModifierSystem, which means the slowdown will no longer disappear 3 seconds after receiving damage, and will not stack (however, it now scales with stamina damage).
- When you exit stamcrit, you start at (100 - epsilon) stamina damage. There still exists another check that prevents you from getting re-stunned in the next ~5 seconds (maybe we should remove that too?), but the slowdown and combat disadvantages will apply as they are supposed to. This means you can no longer stand up after being exhausted to the point of fainting on the ground and immediately rush back into combat.

In addition to that, I also did the following:
- Re-added the mass contest to EscapeInventorySystem. It seems like it's been nuked when the new mass contest system was being implemented and never added back.
- Fixed the mass contest in carrying again

<p>

https://github.com/user-attachments/assets/076b3c3b-cdd1-4ec7-969d-2564c814a40e

</p>
</details>

---

# Changelog
:cl:
- fix: Shoving once again works correctly, and mass difference matters a lot when shoving someone.
- fix: The time it takes to escape one's hands once again depends on the mass difference between the escapee and the holder.
- tweak: Exiting stamina crit now leaves you with 0 stamina. You can't be immediately stunned again, but you will suffer from slowdown and combat disadvantages!

### DIFF
--- a/Content.Server/Carrying/CarryingSystem.cs
+++ b/Content.Server/Carrying/CarryingSystem.cs
@@ -183,7 +183,10 @@ namespace Content.Server.Carrying
             // Check if the victim is in any way incapacitated, and if not make an escape attempt.
             // Escape time scales with the inverse of a mass contest. Being lighter makes escape harder.
             if (_actionBlockerSystem.CanInteract(uid, component.Carrier))
-                _escapeInventorySystem.AttemptEscape(uid, component.Carrier, escape, _contests.MassContest(uid, component.Carrier, false, 2f));
+            {
+                var disadvantage = _contests.MassContest(component.Carrier, uid, false, 2f);
+                _escapeInventorySystem.AttemptEscape(uid, component.Carrier, escape, disadvantage);
+            }
         }
 
         private void OnMoveAttempt(EntityUid uid, BeingCarriedComponent component, UpdateCanMoveEvent args)
--- a/Content.Server/Resist/EscapeInventorySystem.cs
+++ b/Content.Server/Resist/EscapeInventorySystem.cs
@@ -69,8 +69,8 @@ public sealed class EscapeInventorySystem : EntitySystem
         // Contested
         if (_handsSystem.IsHolding(container.Owner, uid, out _))
         {
-            var advantage = _contests.MassContest(uid, container.Owner, rangeFactor: 3f);
-            AttemptEscape(uid, container.Owner, component, advantage);
+            var disadvantage = _contests.MassContest(container.Owner, uid, rangeFactor: 3f);
+            AttemptEscape(uid, container.Owner, component, disadvantage);
             return;
         }
 
--- a/Content.Server/Resist/EscapeInventorySystem.cs
+++ b/Content.Server/Resist/EscapeInventorySystem.cs
@@ -7,6 +7,7 @@ using Content.Shared.Hands.EntitySystems;
 using Content.Server.Storage.Components;
 using Content.Shared.ActionBlocker;
 using Content.Shared.Actions;
+using Content.Shared.Contests;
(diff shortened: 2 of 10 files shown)


Content.Server/Carrying/CarryingSystem.cs
```diff
@@ -183,7 +183,10 @@ private void OnMoveInput(EntityUid uid, BeingCarriedComponent component, ref Mov
             // Check if the victim is in any way incapacitated, and if not make an escape attempt.
             // Escape time scales with the inverse of a mass contest. Being lighter makes escape harder.
             if (_actionBlockerSystem.CanInteract(uid, component.Carrier))
-                _escapeInventorySystem.AttemptEscape(uid, component.Carrier, escape, _contests.MassContest(uid, component.Carrier, false, 2f));
+            {
+                var disadvantage = _contests.MassContest(component.Carrier, uid, false, 2f);
+                _escapeInventorySystem.AttemptEscape(uid, component.Carrier, escape, disadvantage);
+            }
```
VMSolidus, line 189:
```suggestion
                _escapeInventorySystem.AttemptEscape(uid, component.Carrier, escape, _contests.MassContest(component.Carrier, uid, false, 2f));
```

Content.Server/Resist/EscapeInventorySystem.cs
```diff
@@ -67,7 +69,8 @@ private void OnRelayMovement(EntityUid uid, CanEscapeInventoryComponent componen
         // Contested
         if (_handsSystem.IsHolding(container.Owner, uid, out _))
         {
-            AttemptEscape(uid, container.Owner, component);
+            var disadvantage = _contests.MassContest(container.Owner, uid, rangeFactor: 3f);
+            AttemptEscape(uid, container.Owner, component, disadvantage);
```
VMSolidus, line 73:
```suggestion
            AttemptEscape(uid, container.Owner, component, _contests.MassContest(container.Owner, uid, rangeFactor: 3f));
```

VMSolidus (changes requested):

Content.Server/Resist/EscapeInventorySystem.cs
Mnemotechnician: No, this makes it clear this is DISADVANTAGE and not the opposite. It's best for maintainability to keep it separate, because I had to fix this exact mistake four times across three forks: once on frontier, once on delta, before contests rework on EE, and now finally post-rework on EE.

VMSolidus: Fine, whatever.

VMSolidus (approved):

OldDanceJacket (approved):

PR #814, opened 2024-08-28
Americanize Contests System

VMSolidus:
# Description

I have updated the Contests System to make it safe to use by Americans. This is accomplished by adding very helpful warning labels, as well as restricting the Range of all its functions so that they cannot return either a negative number, or a zero. Therefore, all Contests are now perfectly safe to divide by. 

# Changelog

:cl:
- tweak: ContestsSystem(Mass Contest, Stamina Contest, Health Contest, Mind Contest, Mood Contest, and Every Contest) have all been updated to be safe to divide by, since it is no longer possible for them to ever return 0. Instead they will just return 0.0000000000000000000000000000000000000000000001401298, so if you divide by an unclamped contest, I hope you like your function randomly returning the 32bit integer limit.
- add: Warning labels have been added to all Contests to make them comply with the demands of the American consumer market.


Content.Shared/Contests/ContestsSystem.cs
```diff
@@ -38,11 +41,12 @@ public float MassContest(EntityUid performerUid, bool bypassClamp = false, float
...
                 return 1f;
 
-            return _cfg.GetCVar(CCVars.AllowClampOverride) && bypassClamp
+            return Math.Clamp(_cfg.GetCVar(CCVars.AllowClampOverride) && bypassClamp
                 ? performerPhysics.Mass / otherMass
                 : Math.Clamp(performerPhysics.Mass / otherMass,
                     1 - _cfg.GetCVar(CCVars.MassContestsMaxPercentage) * rangeFactor,
-                    1 + _cfg.GetCVar(CCVars.MassContestsMaxPercentage) * rangeFactor);
+                    1 + _cfg.GetCVar(CCVars.MassContestsMaxPercentage) * rangeFactor),
+                    float.Epsilon, float.MaxValue);
```
Remuchi, line 49: you should probably extract this code part in the separate method and replace all the FOURTEEN occurrences of it with this method

Elijahrane:
Are changelogs supposed to be player-facing or not

Content.Shared/Contests/ContestsSystem.cs
VMSolidus: Got it.

DEATHB4DEFEAT (approved):
There's a XMLDoc tag for arguments, you can move the remarks about the `bypassClamp` to those if you want.

OldDanceJacket (approved):

PR #816, opened 2024-08-29
Programmable Logic Controller

Timfa2112:
# Description
A PLC, or Programmable Logic Controller, which interfaces with the wireless Device Linking system.
Build the PLC, write Spessembly code and have your PLC control and read from any attached devices.

You could use these PLC's to add dynamic behaviour to anything that you can link together with your multitool. Think airlocks that actually pump in or out air before cycling, more complex behaviour in conveyor belts, longer sequences of events when pressing buttons or levers, automating linked devices based on pressure, temperature, time, light intensity, etc. etc.

---

# TODO

- [x]  C# class that can interpret assembly-like code
- [x]  defining input and output registers in the assembly-interpreter
- [x]  complete assembly-like language (working title [Spessembly](https://gitlab.com/Timfa/spessembly)) 
- [ ] Creating the boilerplate code needed for the entity to exist and do things in the world
- [ ]  connecting Device Linking signals and sending integers over the connection for sensors 
- [ ]  Sprites! 
- [x]  Buildable object via build menu, something simple like steel and lv cable 
- [ ]  GUI the user can type text in, which also displays console output
- [ ]  dynamically adding device linking inputs and outputs 

---

<p>

Example Media Embed

</p>
</details>

---

# Changelog

:cl: Alia Ash
- add: Added Programmable Logic Controller


BasedUser:
<details><summary>no way mlog ss14</summary>

image

</details>

Content.Shared/PLC/Systems/PLCInterpreter.cs
```diff
@@ -0,0 +1,438 @@
...
+        private const char pinPrefix = 'p';
+        private const char registerPrefix = 'r';
+        private const char immediatePrefix = '#';
+        private const char addressStart = '[';
+        private const char addressEnd = ']';
+
+        public const int HIGH = 100;
+        public const int LOW = 0;
+
+        private string[] code = Array.Empty<string>();
```
BasedUser, line 20: ideally these should be non-string as string comparisons take a LOT of computing power, something like:
```cs
private string code = "";
private string instructions = Array.Empty<PLCInstruction>();

// ...

public class PLCInstruction
{
    // yml hooks??? idfk maybe i'm getting WAY too inspired
    public string Name = "Invalid";
    public string Opcode = "nop";
    
    public virtual void Run(PLCVar[] args)
    {
        //Logger.Info("Ran instruction " + self.Name);
        return;
    }
}

public class AddI : PLCInstruction
{
    Name = "Add";
    Opcode = "add";
    
    public override void Run(PLCVar[] args)
    {
        if (args.Length < 3)
            return;
        args[2].SetValue(args[0].Value() + args[1].Value())
        return;
    }
}
```

Timfa2112: the code[] array is the actual text typed by the player, so I'd need to load it all in and then convert it to a separate array with instructions. Still have to check the strings, but then I do it only once instead of every step.
Probably a good idea, but I'm going to focus on getting it to work, and then optimising it after a working version is achieved

BasedUser: optionally throw NMIs instead of returning void lmao

Resources/Prototypes/Entities/Structures/Machines/plc.yml
```diff
@@ -0,0 +1,47 @@
+- type: entity
+  id: PLC
+  name: plc
+  description: It's a programmable logic controller.
```
Mnemotechnician, line 4: This should inherit from BaseStructure. Transform, Clickable, TnteractionOutline, Damageable will all be unnecessary as they are defined in BaseStructure.

Alternatively, you could create a machine board for this and make it inherit from BaseMachine - in that case you'll need to look into how other machines (e.g. autolathe) do it.

Mnemotechnician (changes requested):

Resources/Prototypes/Entities/Structures/Machines/plc.yml
Timfa2112: My intention is to keep it simple.
How do I inherit from BaseStructure? Do I just remove those types under components and replace them with one BaseStructure type?

Mnemotechnician: Uh, add `parent: BaseStructure`. Then navigate to the BaseStructure type, look what components it already has that were re-defined in your type, and remove them. If the parent entity defines a component but does not define all the fields you want, you can still add the same component and only set up the fields that were not set up in the parent entity. In that case, The rest of the fields will be inherited from the parent.

Timfa2112:
> Is this being worked on?

I try to figure things out now and then, but I haven't had time in the past few days.
Because I'm still very unfamiliar with how the device linking works, as well as making GUI and ensuring the entity to be placed in the world functions properly, I've attempted to find someone who wants to collaborate on this and fill in the gaps of my knowledge, but I've had no luck with that yet.

So, in short, it's had no progress on a while, but I haven't given up yet :P

PR #853, opened 2024-09-04
Fix

Evgencheg:
this might fix bug with clothes that are not updatable after pressing button


VMSolidus:
This has already been fixed by Loadouts Version 3.

PR #856, opened 2024-09-04
Disable Quick Swap for Item Slots by Default

Mnemotechnician:
# Description
As was discussed on discord, quick swap is a feature that makes certain things needlessly easy. You can instantly swap the magazine in a gun without even paying attention to it, you can quickly swap cells in IPCs/borgs, you can instantly swap the ID in your pda with that of someone else without anyone noticing, et cetera. In my opinion, this is unnecessary QOL that makes the game feel more bland and primitive. Besides, it used to lead to odd and unwanted behavior, such as IPCs being able to swap their own power cells.

In addition to that, certain entities do not support quick swap. For instance, rechargers will stop working if you insert a power cell and then quick-swap it before it finishes charging.

This PR makes quick swap disabled by default. It can still be re-enabled by setting `ItemSlotsComponent.Slots.[ItemSlot].Swap = true` on individual entities.

# Changelog
:cl:
- remove: Quick swap has been disabled for most item slots. This primarily means you will have to eject power cells/magazines from items/weapons/borgs before replacing them with different ones.


Vaaankas:
having to actually use hands to interact with things ??? crazy !

gluesniffler:
make it a cvar

Mnemotechnician:
> make it a cvar

Not sure how to achieve that since the assignment happens in the primary constructor of the class.

DEATHB4DEFEAT (approved):

DEATHB4DEFEAT:
> > make it a cvar
> 
> Not sure how to achieve that since the assignment happens in the primary constructor of the class.

Modify it on init

Mnemotechnician:
Better solution.

VMSolidus (approved):

PR #858, opened 2024-09-04
Make Height Sliders Affect Your Bloodstream Volume

Mnemotechnician:
# Description
Something that just makes sense, this makes your effect character weight affect your bloodstream volume. As a minimum size felinid you will get 33% of normal blood volume, whereas as something as huge as a lamia you may get up to 3 times the normal blood volume.

The resulting volume of your bloodstream can be calculated as `V = clamp(normal_volume * mass_contest ^ 0.6)` (assuming default parameters), where mass_contest is the result of a mass contest between your entity and the average humanoid. For average species like vulps, this means that their bloodstream can become up to ~40% smaller than normal (at minimum size), or up to 50% larger than normal (at maximum size). For onis the range is shifted towards higher values, a maximum size oni will have twice as much blood as an average human.

This has both drawbacks and advantages. For instance, having little blood means you can bleed out easily, but at the same time it means it will take way less blood packs/saline/iron/proteins to restore your blood to the normal level. Opposite is also true, having more blood means you will be harder to heal.

Also, this PR slightly refactors the HeightAdjustSystem to be more flexible.

<p>

https://github.com/user-attachments/assets/951c2391-09d8-4a4a-812b-a2394862fadd

</p>
</details>

# Changelog
:cl:
- add: Your character size now affects your blood level. Smaller characters will have less blood, and larger characters will have more.

### DIFF
--- a/Content.Server/HeightAdjust/BloodstreamAdjustSystem.cs
+++ b/Content.Server/HeightAdjust/BloodstreamAdjustSystem.cs
@@ -1,0 +1,45 @@
+using Content.Server.Body.Components;
+using Content.Server.Chemistry.Containers.EntitySystems;
+using Content.Shared.CCVar;
+using Content.Shared.Chemistry.Reagent;
+using Content.Shared.Contests;
+using Content.Shared.HeightAdjust;
+using Robust.Shared.Configuration;
+
+namespace Content.Server.HeightAdjust;
+
+public sealed class BloodstreamAdjustSystem : EntitySystem
+{
+    [Dependency] private readonly IConfigurationManager _config = default!;
+    [Dependency] private readonly ContestsSystem _contests = default!;
+    [Dependency] private readonly SolutionContainerSystem _solutionContainer = default!;
+
+    public override void Initialize()
+    {
+        SubscribeLocalEvent<BloodstreamAffectedByMassComponent, MapInitEvent>((uid, comp, _) => TryAdjustBloodstream((uid, comp)));
+        SubscribeLocalEvent<BloodstreamAffectedByMassComponent, HeightAdjustedEvent>((uid, comp, _) => TryAdjustBloodstream((uid, comp)));
+    }
+
+    /// <summary>
+    ///     Adjusts the bloodstream of the specified entity based on the settings provided by the component.
+    /// </summary>
+    public bool TryAdjustBloodstream(Entity<BloodstreamAffectedByMassComponent> ent)
+    {
+        if (!TryComp<BloodstreamComponent>(ent, out var bloodstream)
+            || !_solutionContainer.TryGetSolution(ent.Owner, bloodstream.BloodSolutionName, out var bloodSolutionEnt)
+            || !_config.GetCVar(CCVars.HeightAdjustModifiesBloodstream))
+            return false;
+
+        var bloodSolution = bloodSolutionEnt.Value.Comp.Solution;
+
+        var factor = Math.Pow(_contests.MassContest(ent, bypassClamp: true, rangeFactor: 4f), ent.Comp.Power);
+        factor = Math.Clamp(factor, ent.Comp.Min, ent.Comp.Max);
+
+        var newVolume = bloodstream.BloodMaxVolume * factor;
+        var newBloodLevel = bloodSolution.FillFraction * newVolume;
+        bloodSolution.MaxVolume = newVolume;
+        bloodSolution.SetContents([new ReagentQuantity(bloodstream.BloodReagent, newBloodLevel, null)], false);
+
+        return true;
+    }
+}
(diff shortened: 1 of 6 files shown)


VMSolidus (changes requested):
Marking this PR as "Do Not Merge" until https://github.com/Simple-Station/Einstein-Engines/pull/814 is merged, due to this PR depending on an "Unsafe" implementation of Contests, that will require certain safety checks be included in the Contests System.

Content.Server/HeightAdjust/BloodstreamAdjustSystem.cs
```diff
@@ -0,0 +1,45 @@
...
+    public bool TryAdjustBloodstream(Entity<BloodstreamAffectedByMassComponent> ent)
+    {
+        if (!TryComp<BloodstreamComponent>(ent, out var bloodstream)
+            || !_solutionContainer.TryGetSolution(ent.Owner, bloodstream.BloodSolutionName, out var bloodSolutionEnt)
+            || !_config.GetCVar(CCVars.HeightAdjustModifiesBloodstream))
+            return false;
+
+        var bloodSolution = bloodSolutionEnt.Value.Comp.Solution;
+
+        var factor = Math.Pow(_contests.MassContest(ent, bypassClamp: true, rangeFactor: 4f), ent.Comp.Power);
```
VMSolidus, line 35:
```suggestion
        var factor = Math.Pow(_contests.MassContest(ent, bypassClamp: true), ent.Comp.Power);
```
When bypassClamp is set to true, rangeFactor becomes irrelevant, since bypassClamp prevents Range from being a variable, making it so that a given Contest will return (-inf, inf) as its range. This'll be limited by https://github.com/Simple-Station/Einstein-Engines/pull/814 to instead return (Epsilon, 32bit Int Limit).

Content.Server/HeightAdjust/BloodstreamAffectedByMassComponent.cs
```diff
@@ -0,0 +1,26 @@
...
+
+/// <summary>
+///     When applied to a humanoid or any mob, adjusts their blood level based on the mass contest between them
+///     and an average humanoid. <br/><br/>
+///
+///     The formula for the resulting bloodstream volume is <code>V = BloodMaxVolume * MassContest^Power</code>
+///     clamped between the specified Min and Max values.
+/// </summary>
+[RegisterComponent]
+public sealed partial class BloodstreamAffectedByMassComponent : Component
```
VMSolidus, line 13: This could probably be renamed to "MassModifiersComponent", or some variety of such, so that it could be extended into a modular component that handles all MassContest related subscriptions for a variety of systems.

Content.Shared/CCVar/CCVars.cs
```diff
@@ -2294,6 +2294,13 @@ public static readonly CVarDef<float>
         public static readonly CVarDef<bool> HeightAdjustModifiesZoom =
             CVarDef.Create("heightadjust.modifies_zoom", false, CVar.SERVERONLY);
 
+        /// <summary>
+        ///     Whether height & width sliders adjust a player's bloodstream volume.
+        /// </summary>
+        /// <remarks>This can be configured more precisely by modifying BloodstreamAffectedByMassComponent.</remarks>
+        public static readonly CVarDef<bool> HeightAdjustModifiesBloodstream =
+            CVarDef.Create("heightadjust.modifies_bloodstream", true, CVar.SERVERONLY);
```
VMSolidus, line 2302:
```suggestion
        /// <summary>
        ///     Whether height & width sliders adjust a player's bloodstream volume.
        /// </summary>
        /// <remarks>
        ///     This can be configured more precisely by modifying BloodstreamAffectedByMassComponent.
        /// </remarks>
        public static readonly CVarDef<bool> HeightAdjustModifiesBloodstream =
            CVarDef.Create("heightadjust.modifies_bloodstream", true, CVar.SERVERONLY);
```

Content.Server/HeightAdjust/BloodstreamAdjustSystem.cs
Mnemotechnician: It is still relevant in case some genius decides to set AllowClampOverride to true.

Content.Server/HeightAdjust/BloodstreamAffectedByMassComponent.cs
Mnemotechnician: That is the opposite of modular.

VMSolidus: Mmm. Alright.

Content.Server/HeightAdjust/BloodstreamAffectedByMassComponent.cs
```diff
@@ -0,0 +1,26 @@
+using Content.Server.Body.Components;
+
+namespace Content.Server.HeightAdjust;
+
+/// <summary>
+///     When applied to a humanoid or any mob, adjusts their blood level based on the mass contest between them
+///     and an average humanoid. <br/><br/>
+///
```
DEATHB4DEFEAT, line 8:
```suggestion
///     and an average humanoid.
///     <br/>
```

DEATHB4DEFEAT (approved):

PR #861, opened 2024-09-05
[Feature] Shortcut Construction System

Remuchi:
# Description

Interacting with some materials now opens up the radial menu with some often used recipes. Port from White Dream.

---

# Why?

It looks cool I guess. Also required by Blood Cult.

---
<p>

https://github.com/user-attachments/assets/8207ea19-f4a3-45e1-b6f2-c870b46b6837

</p>
</details>

---

# Changelog

:cl:
- add: Interacting with some materials now opens the radial menu containing some of the most used recipes of that material.


Content.Client/ShortConstruction/UI/ShortConstructionMenuBUI.cs
```diff
@@ -0,0 +1,126 @@
...
+        {
+            HorizontalExpand = true,
+            VerticalExpand = true,
+            BackButtonStyleClass="RadialMenuBackButton",
+            CloseButtonStyleClass="RadialMenuCloseButton"
+        };
+
+        var mainContainer = new RadialContainer
+        {
+            Radius = 100
```
Mnemotechnician, line 66: Do not hardcode the radius, or the menu will break like emoji wheel on wizden. Instead, calculate it based on the number of elements. Assuming the menu is a hexagon with n sides and side length 48, its radius should be `48/2 / sin(PI / n / 2)` or `24 / sin(PI / n / 2)`; in this case it would be

```cs
            Radius = 24f / Math.Sin(Math.PI / (crafting.Prototypes.Count - 1) / 2)
```

Though you can set it the base factor higher than 24 to add padding.

Remuchi: Thanks! I was really wondering about the formula for it.

Remuchi: image
Great

Mnemotechnician: Uh, little note... I didn't consider that `crafting.Prototypes.Count - 1` may result in division by 0, and that there's a button in the middle... I guess the `- 1` part should be removed since both are the case.

VMSolidus (approved):
This is the only change I'd request, other than that it's fine.

OldDanceJacket (approved):

PR #884, opened 2024-09-07
Step Trigger Minor Refactor

VMSolidus:
# Description

I made this PR in response to a bug report from Floof, where it was discovered that a "Minimum Size Felinid or Harpy" is instantaneously killed by mouse traps. Which confused me, because Felinids and Harpies are intended to be immune to floor traps like landmines, glass, and mouse traps. Then I discovered that mouse traps overwrite the step trigger cancellation, meaning that mouse traps will just completely ignore a Felinid/Harpy canceling the step trigger. Additionally, to my endless frustration, the Felinid/Harpy floor trap immunity is handled by a Tag and not a Component, which isn't really acceptable in this day and age.

I decided to take a little bit of a different approach to this problem, first by doing the usual code cleanup to EE standards. Then by adding a new StepTriggerImmuneComponent. This component acts as an early-exit for the entire StepTriggerSystem, immediately at the initial entrypoint, during the CanTrigger bool. This component is given to Felinids and Harpies by default, representing their "Extremely low density bodies" having too much surface area and not enough mass to trigger floor traps. Effectively, they are now working as originally intended, by having immunity to setting off landmines.

Because we have a trait point system, and this is coincidentally also a trait requested by Nuclear14, I have gone ahead and created a Trait that gives this component to anyone willing to pay the points for it.

# Changelog

:cl:
- fix: Felinids and Harpies will now correctly never set off floor traps, such as landmines and mouse traps.
- add: Trap Avoider has been added as a new trait, allowing characters to buy the innate ability to avoid floor traps. I would have named this "Light Step", after the trait from Fallout that shares its name and effect, but someone already gave that name to a different trait entirely.


Content.Server/Mousetrap/MousetrapSystem.cs
```diff
@@ -44,15 +40,16 @@ private void OnStepTriggerAttempt(EntityUid uid, MousetrapComponent component, r
...
-            args.Damage *= scaledDamage;
-        }
+        if (!TryComp<PhysicsComponent>(args.Tripper, out var physics)
+            || physics.Mass is 0)
+            return;
+
+        // The idea here is inverse,
+        // Small - big damage,
+        // Large - small damage
+        // yes i punched numbers into a calculator until the graph looked right
```
DEATHB4DEFEAT, line 50: real
```suggestion
        // Yes, I punched numbers into a calculator until the graph looked right
```

Resources/Locale/en-US/traits/traits.ftl
```diff
@@ -216,3 +216,8 @@ trait-description-NaturalTelepath =
     whether or not you possess any notable psychic powers. This offers all of the same benefits and
     drawbacks of Latent Psychic, except that you are guaranteed to start with full Telepathy. You may
     still gain powers as normal for a Latent Psychic.
+
+trait-name-TrapAvoider = Trap Avoider
+trait-description-TrapAvoider =
+    You possess a preturnatural sense of traps, and will unconsciously avoid them. You will never trigger
+    floor traps, such as land mines, tripwires, mouse traps(If you're small enough), etc.
```
DEATHB4DEFEAT, line 223:
```suggestion
    floor traps, such as land mines, tripwires, mouse traps(If you're small enough), etc.

```

Resources/Prototypes/Traits/skills.yml
```diff
@@ -241,3 +241,16 @@
...
+  category: Physical
+  points: -3
+  components:
+    - type: StepTriggerImmune
+  requirements:
+    - !type:CharacterSpeciesRequirement
+      inverted: true
+      species:
+        - Felinid
+        - Harpy
```
DEATHB4DEFEAT, line 256:
```suggestion
        - Harpy

```

DEATHB4DEFEAT (approved):

OldDanceJacket (approved):

PR #899, opened 2024-09-10
Non-Destructive Universal Language

VMSolidus:
# Description

UniversalLanguageSpeakerComponent has NUMEROUS issues, which previously were easy to ignore since it was a language only obtainable by Admins. But now that it's both a Psionic Power(And a trait that adds said power), the issues with it have been highlighted. Here's just a FEW

1. UniversalLanguageSpeaker overwrites all known languages, preventing you from speaking anything else
2. It overwrites the ability to use sign language.
3. It also overwrites the Mute trait.

To fix that, I've made it follow *MOSTLY* the logic all the other languages use, so that it's less of a special snowflake case. Now if you have access to it, it will appear in your language list alongside other languages, rather than fully replacing the entire list. That way you can intentionally choose not to speak in a language understood by all.

Fuck it, I also added the ability for the TraitSystem to just call LanguageSystem and directly add arbitrarily any desired language, rather than needing a component to do so.

# Changelog

:cl:
- fix: UniversalLanguageSpeaker(And Xenoglossy by extension) will now appear in your language menu alongside other known languages, rather than replace all known languages. You can effectively now choose whether or not to speak it, or to use a normal language.
- add: Traits can now add Languages directly.
- add: Traits can now remove Languages directly.


FoxxoTrystan:
Reading this from a laptop, i suggest letting @Mnemotechnician a read as currently unable to do it correctly.

Alithsko:
I'm the one who brought this to light to OP, and it was worse than overwriting mute as it gave you the message that you cannot speak with no other mute friendly choices like sign, which made it a strict downgrade if you get it, happened to me in a game so this is a godsend

Mnemotechnician (changes requested):
On a side note, I hate how this PR is straight up hardcoding this behavior in the trait system, but I guess it's acceptable for now since we do not have a proper system in place... Maybe in the future we can refactor it to accept polymorphic actions instead, similarly to e.g. reagent effects, or damage behaviors, or interactions...

VMSolidus:
https://github.com/user-attachments/assets/4dbdfff5-d41a-468c-8229-94e29d75dac1

I have finished fixing this PR, and finished testing to verify that everything works correctly.

VMSolidus:
I ALSO REQUIRE THIS PR IN ORDER TO CONTINUE WORKING ON PSIONICS.

Content.Server/Language/LanguageSystem.cs
```diff
@@ -16,8 +15,18 @@ public override void Initialize()
...
 
         SubscribeLocalEvent<LanguageSpeakerComponent, ComponentInit>(OnInitLanguageSpeaker);
+        SubscribeLocalEvent<UniversalLanguageSpeakerComponent, MapInitEvent>(OnUniversalInit);
+        SubscribeLocalEvent<UniversalLanguageSpeakerComponent, ComponentShutdown>(OnUniversalShutdown);
     }
 
+    private void OnUniversalShutdown(EntityUid uid, UniversalLanguageSpeakerComponent component, ComponentShutdown args)
+    {
+        RemoveLanguage(uid, UniversalPrototype);
+    }
```
DEATHB4DEFEAT, line 25:
```suggestion
    }

```

DEATHB4DEFEAT (approved):

VMSolidus:
image

Translator on: 
image

Translator off:
image

Done here,

Content.Server/Language/LanguageSystem.cs
```diff
@@ -129,20 +129,18 @@ public void RemoveLanguage(
         EntityUid uid,
         string language,
         bool removeSpoken = true,
-        bool removeUnderstood = true,
-        LanguageKnowledgeComponent? knowledge = null,
-        LanguageSpeakerComponent? speaker = null)
```
Mnemotechnician, line 134: Do not remove those. If it bothers you so much, make them into `Entity<LanguageSpeakerComponent?, LanguageKnowledgeComponent?>`.

Content.Server/Language/LanguageSystem.cs
```diff
@@ -168,9 +166,9 @@ public bool EnsureValidLanguage(EntityUid entity, LanguageSpeakerComponent? comp
     /// <summary>
     ///     Immediately refreshes the cached lists of spoken and understood languages for the given entity.
     /// </summary>
-    public void UpdateEntityLanguages(EntityUid entity, LanguageSpeakerComponent? languages = null)
```
Mnemotechnician, line 171: Same here.

Mnemotechnician (changes requested):
Do not remove the optional component parameters. Passing components around instead of fetching them anew every time they are needed is the preferred style in ss14. You can use `Entity<SomeComp>` instead of EntityUid if you want prettier code.

VMSolidus:
> Do not remove the optional component parameters. Passing components around instead of fetching them anew every time they are needed is the preferred style in ss14. You can use `Entity<SomeComp>` instead of EntityUid if you want prettier code.

No.

FoxxoTrystan (approved):
Could have been nicer then just "no"
But at least this works.

VMSolidus:
> Could have been nicer then just "no" But at least this works.

Yep, and I'm going to go immediately start work on Language Menu now.

PR #954, opened 2024-09-22
Fix HealthContest Crash

VMSolidus:
# Description

Apparently _mobThreshold.TryGetThresholdForState() fucking crashes if you hand it a mob that doesn't have a MobStateComponent, even though it optionally accepts no MobStateComponent at all as one of its inputs. Easy solution for this is to just pre-emptively TryComp for the MobStateComponent and then pass it in explicitly, but I also hate this because TryGetThresholdForState shouldn't be crashing if you hand it a mob without a MobStateComponent... I hate Resolve() so much for this behavior. 

# Changelog

:cl:
- fix: Fix a crash related to Luminous Entities.

### DIFF
--- a/Content.Shared/Contests/ContestsSystem.cs
+++ b/Content.Shared/Contests/ContestsSystem.cs
@@ -2,6 +2,7 @@ using Content.Shared.Abilities.Psionics;
 using Content.Shared.CCVar;
 using Content.Shared.Damage;
 using Content.Shared.Damage.Components;
+using Content.Shared.Mobs.Components;
 using Content.Shared.Mobs.Systems;
 using Content.Shared.Mood;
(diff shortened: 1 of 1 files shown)


DEATHB4DEFEAT:
> Apparently _mobThreshold.TryGetThresholdForState() fucking crashes if you hand it a mob that doesn't have a MobStateComponent, even though it optionally accepts no MobStateComponent at all as one of its inputs

Resolve is used to get a component something is expected to have or use the provided one, the function's field is optional because it's more convenient to not get the component for the function if you don't already have it.

FoxxoTrystan (approved):

DEATHB4DEFEAT (approved):

PR #1062, opened 2024-10-16
Adds TC To Listening Post Operatives

gluesniffler:
# Description

Spiritual successor to an old unintended interaction in Delta V, where LP ops would spawn with uplinks. Except instead of having objectives its just an uplink with 20TC.

Why? Because it turns LP Ops from a simple nuisance into a potential threat that can send support on station for you, and makes taking a trip to the outpost to cash in some TC a very viable strategy. Salvage might be more inclined to raid them too, so thats gonna be fun.

# Changelog

:cl: Mocho
- add: Added an uplink with 20TC for all Listening Post Operatives


VMSolidus (approved):

FoxxoTrystan (approved):
Personnaly not a fan of this, i prefer making just the loot on-site much better but this works.

DEATHB4DEFEAT (approved):

PR #1081, opened 2024-10-17
Fix Cryptobiolin

### LINKED ISSUE #1080: Cryptobiolin does not appear to insulate people as it claims to.
# Description

I have been testing cryptobiolin extensively on myself and on other characters, but it does not appear to block any psionic abilities as advertised - telepathic messages still can be both sent and received and the character still appears to be influenced by the abilities that should not influence someone properly psionically insulated.

# Reproduction

1. Apply cryptobiolin to a character such that they are metabolizing it, whether through drinking it, taking a pill, or injecting it. (I can only assume that the affected speech and vision should apply at the same time as the insulation.)
2. Have that character attempt to speak through telepathy if it is available, or otherwise use a psionic ability that should not work on an insulated character on them. (The abilities that I personally tested were mind swap and healing word - the former I had attempted to use on a character wearing an insulated skullcap, and it did not work. When the character was actively metabolizing cryptobiolin, however, both of them worked.)

# Expected behavior

1. Apply cryptobiolin to a character such that they are metabolizing it, whether through drinking it, taking a pill, or injecting it.
2. The character should not be able to send or receive telepathic messages, and abilities that should not affect a psionically insulated character will not affect them.

# Media

In this screenshot, I show off that my speech is scrambled due to cryptobiolin's influence, as well as the drunken visual filter, but I can still send and receive telepathic messages as well as use the shade-skip ability. I tested with an insulated skullcap and was unable to use the power.
cryptobiolintest

# Additional context

VMSolidus:
# Description

Closes #1080
This PR fixes Cryptobiolin so that it correctly adds a temporary "PsionicsDisabled" and "PsionicallyInsulated" status effect to the consumer. This effect lasts for 15 minutes, and is intended to be a highly effective means of providing protection from psionics, especially in the case of a Code White call.

# Changelog

:cl:
- fix: Fixed Cryptobiolin not providing temporary Psionic Insulation. It now provides protection from psionic abilities and events for 15 minutes.

### DIFF
--- a/Resources/Prototypes/Entities/Mobs/Species/base.yml
+++ b/Resources/Prototypes/Entities/Mobs/Species/base.yml
@@ -141,8 +141,8 @@
     - TemporaryBlindness
     - Pacified
     - StaminaModifier
-    - PsionicsDisabled #Nyano - Summary: PCs can have psionics disabled.
-    - PsionicallyInsulated #Nyano - Summary: PCs can be made insulated from psionic powers.
+    - PsionicsDisabled
+    - PsionicallyInsulated
   - type: Reflect
     enabled: false
     reflectProb: 0
@@ -273,6 +273,8 @@
     - Muted
     - Pacified
     - StaminaModifier
+    - PsionicsDisabled
+    - PsionicallyInsulated
   - type: Blindable
   # Other
   - type: Temperature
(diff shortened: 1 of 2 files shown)


FoxxoTrystan (approved):

FoxxoTrystan:
In details what does PsionicsDisabled does exactly?

DEATHB4DEFEAT (approved):

DEATHB4DEFEAT:
> In details what does PsionicsDisabled does exactly?

Can't use your own powers

PR #1110, opened 2024-10-20
Port Sacrificing

VMSolidus:
# Description

This PR rebases the Sacrificing mechanic from Nyanotrasen, while updating it to newer standards for the Reworked Psionics System. This does not rebase Golemancy, or Items Of Power, so the system for now exists solely to spike glimmer DOWN in the event of a code white, and merely produces brains to be placed inside cyborgs. Notably different from DeltaV's implementation of it, this system does not reduce a random amount of glimmer, and instead reduces glimmer by a base of 25, multiplied by the victim's Psionic Amplification(which is itself semi-random). Which means that certain individuals(Especially people with lots of powers, and/or the Power Overwhelming trait) are inherently more useful to sacrifice than other, weaker psionics. 

When I bring back Items Of Power, the chances of obtaining an Artifact are based on the victim's Dampening stat.

https://github.com/user-attachments/assets/23e52f20-ac3e-4bd9-8122-cadec96cbdf3

# Changelog

:cl:
- add: Sacrificing Psions has been added. Psions can be sacrificed by Epistemics upon an altar in order to dramatically reduce glimmer. The more powerful the Psion to be sacrificed, the more glimmer is reduced.


Content.Server/Chat/Systems/ChatSystem.cs
FoxxoTrystan, line 1: Is this supose to be here?

FoxxoTrystan (changes requested):

Content.Server/Chat/Systems/ChatSystem.cs
VMSolidus: Yes. I had to revert it to test. This will go away when I hit merge master.

Peptide90 (approved):

FoxxoTrystan (approved):

Issue #1101, opened 2024-10-19
Custom Species Name disappears after switching race, when disabled in the prototype


DangerRevolution:
# Reproduction

Enter the Character UI
Swap race

# Expected behavior

It doesn't do that

# Media

https://github.com/user-attachments/assets/801c1f74-3d3b-4f20-8760-715cf315e688

This happens whether or not you input any text.
This does not return if you restart the round, nor re-enter the lobby. You have to full restart to get the menu back.

DangerRevolution:
Yes, this was also a problem in [dde6130](https://github.com/Simple-Station/Einstein-Engines/commit/dde61304ff56a54844f599758e8b84d95ee9f0fa) (commit after #994). Not sure how this passed review given this is findable within 5 seconds of interacting with the menu.

FoxxoTrystan:
> # Reproduction
> Enter the Character UI Swap race
> 
> # Expected behavior
> It doesn't do that
> 
> # Media
>  2024-10-19.22-25-07.mp4 
> This happens whether or not you input any text. This does not return if you restart the round, nor re-enter the lobby. You have to full restart to get the menu back.

Custom Specie Name menu should be visible to begin with, as goob MRP (if i think this is where its from) did not enable those.

The name should not show up actully unless enabled, did you enable the CustomName variable?

DangerRevolution:
Tested on the latest commit: [d7de808](https://github.com/Simple-Station/Einstein-Engines/commit/d7de8082d3b3c77832a2c190db1e0c03752cb930)

> Custom Specie Name menu should be visible to begin with, as goob MRP (if i think this is where its from) did not enable those.

A) As I said, the screenshot is from the [dde6130](https://github.com/Simple-Station/Einstein-Engines/commit/dde61304ff56a54844f599758e8b84d95ee9f0fa) commit; not the Goob playtest?

> The name should not show up actully unless enabled, did you enable the CustomName variable?

B) The issues are still exactly the same as the video regardless of the prototype:
i) Enabling the prototype still means the custom name field disappears in-between switching
ii) Using the field pre-switch doesn't make the name appear in-game
ii a) Using the name field pre-switch and saving the character slot does not make it appear does not make the name show in-game
ii b) Using the name field pre-switch and changing another variable and saving the character slot does not make the name show in-game
iii) If you type a custom species name into slot A; then switch to slot B then the custom name carries over

# Media

[...]

PR #1120, opened 2024-10-21
Fix Custom Specie Name

Aidenkrz:
# Description

Custom specie name wouldn't properly update when a different race was selected, or when you changed profiles.
This fixes that.
Also makes it so the text field is automatically filled with the default name instead of an empty space.
(and makes the species check less bad 😺 )

Solves #1101

---

<p>

https://gyazo.com/63ba2d8a3456c19c70f25164ae42078d

</p>
</details>

---

# Changelog

:cl:
- fix: Custom specie name doesn't disappear in the editor anymore.


FoxxoTrystan (approved):
I love wizden PR, making a mess, good job tho

VMSolidus (approved):
I love string fuckery so much. Wack.

PR #1147, opened 2024-10-26
Voice Mask Refactor and Fixes

beck-thompson:
Based off [Voice Mask refactor](https://github.com/space-wizards/space-station-14/pull/30798) (The license is also different there so if need a different one, use that PR). Had to make some changes because of the implants but overall its basically the same!

Fixed at least 2 bugs:
You can now use the voice mask.
You can now use the harpy voice mask implant properly.

Codebase changes:
The `VoiceMaskComponent` should not be used for masking an entities voice. It is now used for the voice mask item!

If you were using `VoiceMaskComponent` to change an entities voice, instead listen for the `TransformSpeakerSpeechEvent` or use the `VoiceOverrideComponent` (If possible use the event!).

If your system was adding the `VoiceMaskComponent` when an item was equipped and then removing it when unequipped use the `InventoryRelayedEvent<TransformSpeakerSpeechEvent>` instead! The `TransformSpeakerSpeechEvent` is relayed to worn clothing items now.

Also, you can and should now remove all `VoiceMaskBoundUserInterface`s from entities that use voice masks. The interface is now stored on the voice mask itself.

<p>

https://github.com/user-attachments/assets/0835e720-f8f7-49ad-beb3-aa77d14225ec

</p>
</details>

---

# Changelog

:cl: Beck Thompson
- fix: Fixed various bugs related to voice masks and voice mask implants.


beck-thompson:
Also, I'm not that familiar with EE so if could have easily missed some common voice editing thing. Probably test this a little more than normal 😆

For testing I'd recommend checking:
1.) Voice mask itself (Duh)
2.) The bionic syrinx implant.
3.) Combining the implant and the mask
4.) The intercom hacking ability (Pulsing that one wire)

VMSolidus (approved):
Oh, this is just cherrypicking a wizden refector, that's completely fine. For reference we can take anything we want from Wizden, their license literally tells us we can.

OldDanceJacket (approved):

PR #1157, opened 2024-10-28
Port Cool Punpun

DEATHB4DEFEAT:
https://github.com/Simple-Station/Parkstation/pull/48
https://github.com/Simple-Station/Parkstation-Friendly-Chainsaw/issues/2


Content.Server/Punpun/PunpunSystem.cs
```diff
@@ -0,0 +1,109 @@
...
+        { 5, "VI" },
+        { 6, "VII" },
+        { 7, "VIII" },
+        { 8, "IX" },
+        { 9, "X" },
+        { 10, "XI" },
+        { 11, "XII" },
+        { 12, "XIII" },
+        { 13, "XIV" }
+    };
```
VMSolidus, line 35:
```suggestion
    // All the roman numerals we'll need to display
    private readonly Dictionary<int, string> _numerals = new()
    {
        { 0, "I" },
        { 1, "II" },
        { 2, "III" },
        { 3, "IV" },
        { 4, "V" },
        { 5, "VI" },
        { 6, "VII" },
        { 7, "VIII" },
        { 8, "IX" },
        { 9, "X" },
        { 10, "XI" },
        { 11, "XII" },
        { 12, "XIII" },
        { 13, "XIV" }
    };
```
Do you think we can port in a roman numeral constructor here? I have a specific reason for asking.

Content.Server/Punpun/PunpunSystem.cs
```diff
@@ -0,0 +1,109 @@
...
+        base.Initialize();
+
+        SubscribeLocalEvent<PunpunComponent, ComponentStartup>(OnRoundStart);
+        SubscribeLocalEvent<RoundEndTextAppendEvent>(OnRoundEnd);
+    }
+
+    // Checks if the Punpun data has any items to equip, and names the Punpun upon initialization
+    private void OnRoundStart(EntityUid uid, PunpunComponent component, ComponentStartup args)
+    {
+        if (_punpunData.Item1 > 13)
```
VMSolidus, line 48:
```suggestion
        if (_punpunData.Item1 > component.MaxLives)
```
And the reason I'm asking about porting in a constructor, is that I actually think this 13 here should be instead a variable, because someone might hypothetically want to set it to some absurdly high number. :)

That and 13 on its is a "Magic Number" that breaks our own maintainer standards. It should either be a constant, or better yet a component variable.

Content.Server/Punpun/PunpunComponent.cs
```diff
@@ -0,0 +1,4 @@
+namespace Content.Server.Punpun;
+
+[RegisterComponent]
+public sealed partial class PunpunComponent : Component;
```
VMSolidus, line 4: I had no idea you could just close a component with a semicolon like that, I thought you had to close them with { }
Huh.

VMSolidus (changes requested):

Content.Server/Punpun/PunpunSystem.cs
DEATHB4DEFEAT: I was thinking about that, I can look for something

Content.Server/Punpun/PunpunSystem.cs
```diff
@@ -0,0 +1,114 @@
...
+using Robust.Server.GameObjects;
+
+namespace Content.Server.Punpun;
+
+public sealed class PunpunSystem : EntitySystem
+{
+    [Dependency] private readonly InventorySystem _inventory = default!;
+    [Dependency] private readonly ServerMetaDataSystem _meta = default!;
+
+    private (int, string, string, string) _punpunData = (1, string.Empty, string.Empty, string.Empty);
```
Mnemotechnician, line 16: This should really be a named struct/class

VMSolidus (approved):
This is fine.

Peptide90 (approved):

Issue #1130, opened 2024-10-22
Italicize text when whispering


SkippyMcSkips:
# Description

Currently, whispering's only indication is in the chat window. This feels clunky to me and could be improved by italicizing and darkening the font both over the player's head, and the whole text string in the chat window. It's currently only italicizing the player's name and the word "whispers" in the chat window.

# Media

Example on Goob Station 1, showing how whispering is handled
<img width="251" alt="whispering" src="https://github.com/user-attachments/assets/e651f6b7-b84e-4cf1-917f-20f63e124eac">

Example on EE-based Goob Station 2 (MRP Edition)
![image](https://github.com/user-attachments/assets/6033f9ba-99d7-4478-bc68-e382b2d5b274)

![image](https://github.com/user-attachments/assets/44776d72-e258-42db-a4df-7a7849b8b419)

angelofallars:
Fixed with https://github.com/Simple-Station/Einstein-Engines/pull/1294

PR #1161, opened 2024-10-28
Tweak Whisper Display

PlasmaRaptor:
Tweak Issue #1130 - Make whisper display darker and italic, mimicking Goon Station 1 as requested.


Mnemotechnician:
1. This hardcodes the message color, as opposed to the general consensus of "language colors should be displayed in whispers". Instead of doing that, you should modify the code to dim the color instead.
2. Please, follow the PR template format. Or at least remove the unnecessary parts of it.

PlasmaRaptor:
> 1. This hardcodes the message color, as opposed to the general consensus of "language colors should be displayed in whispers". Instead of doing that, you should modify the code to dim the color.
> 2. Please, follow the PR template format. Or at least remove the unnecessary parts of it.

Ok, I will try to do this and resubmit it.

Mnemotechnician:
You can edit this PR instead, no need to re-submit.

PlasmaRaptor:
I didn't know that.

PlasmaRaptor:
I have edited it. 

*It changes the text style to italic, making it much easier to notice.
*It keeps the color text for language, so you can easily identify the language.

This is not exactly what he requested but it is a big step in the right direction and hopefully will make his whispering experience pleasant. I could not dim the color after trying for a few hours.

Aidenkrz:
Was this actually tested? I don't believe [italic] would work as italicized fonts are seperate fonts, no?

Mnemotechnician:
[italic] will override the [font] tag - so yes, it will KINDA work, but not as intended.

PlasmaRaptor:
I consider it a feature that will differentiate it even more, but if someone feels like editing it by dimming it and keeping the font the same etc. they're welcome to do it. This is just what I can do to accommodate his request to make whispering more distinguishable.

Mnemotechnician:
Also, there is now have two [italic] tags nested inside one another and a [font] tag that doesn't do anything because it gets overridden.

PlasmaRaptor:
> Also, there is now have two [italic] tags nested inside one another and a [font] tag that doesn't do anything because it gets overridden.

I will quickly remove the extra one.

sleepyyapril:
Given that this had the conflicts fixed before I did it and that it entirely removes language display, consistency is now another issue with this. If you want to re-open this, please add media.

PR #1176, opened 2024-11-02
Syndicate Faction Comes From IFF Tags

VMSolidus:
# Description

This broadly fixes a big with Familiars such that, since they inherit a player's factions, they will attack whoever you are factionally opposed to. Which means if you were a Syndicate Chaplain, you could just summon your bat for free damage. This PR flips this on its head by making it so that Syndicate Agents don't have the Syndicate faction, unless they buy a Syndicate ID. This also means that Traitors cannot enter the Listening Outpost, or Nukie Shuttle, without being fired upon by the turrets, unless they are WEARING an Agent ID. 

However this also adds a cool new interaction for CREW. If you are an NT Crewman, and you get your hands on either an Agent ID or a Syndicate PDA, now you can enter a Listening Post or Nukie Shuttle without being fired upon by the turrets.

# Changelog

:cl:
- add: Syndicate Agents no longer have the Syndicate faction. Instead, the Syndicate faction comes from your SYNDICATE ID/PDA. This means that if you steal a Syndicate PDA from nukies, and wear it, you won't be shot by their turrets. This also means that syndicate turrets will fire at syndicate agents not wearing an agent ID.
- fix: Fixed Familiars spawned by traitors attacking crew constantly.

### DIFF
--- a/Content.Server/GameTicking/Rules/TraitorRuleSystem.cs
+++ b/Content.Server/GameTicking/Rules/TraitorRuleSystem.cs
@@ -116,9 +116,9 @@ public sealed class TraitorRuleSystem : GameRuleSystem<TraitorRuleComponent>
             Briefing = briefing
         }, mind, true);
 
-        // Change the faction
-        _npcFaction.RemoveFaction(traitor, component.NanoTrasenFaction, false);
-        _npcFaction.AddFaction(traitor, component.SyndicateFaction);
+        // Don't Change the faction, this was stupid.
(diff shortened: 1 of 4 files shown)


bruhmogus:
since it gives the faction does this mean mobs like the footsoldier and shuttle pilot won't shoot crew with a syndicate id

FoxxoTrystan (approved):

OldDanceJacket (approved):

PR #1164, opened 2024-10-29
Loadouts V4

DEATHB4DEFEAT:
# Description

# TODO

- [x] Custom name/desc/color tint, toggleable individually per-loadout
	- [x] Fix them not changing correctly between profiles in-editor
	- [x] Preview colors in the lobby
	- [x] Allow the users to null the color themselves (and default it to such)
	- [x] Pick what should be allowed to be recolored
- [x] Guidebook links
	- [x] Make an example
- [x] Special components for loadouts
- [x] Heirlooms
	- [x] Pick what should have heirlooms
- [x] Decimate lag
- [x] Fix live character preview
- Maybe do characters per job
  - Rethink unusable

---

<details><summary><h1>Media</h1></summary>
<p>

https://github.com/user-attachments/assets/bcf61517-6b64-40d2-b299-7462e2469fe2

</p>
</details>

---

# Changelog

:cl:
- add: Players can set custom names, descriptions, and color tints for their loadout items
- add: Certain loadouts may have Guidebook pages shown in the editor
- add: Players can pick a list of loadout items to have one randomly be their family heirloom for a mood bonus or deficit if they are carrying it
- fix: Loadouts have almost as little lag as possible (hopefully none)
- fix: Everything properly updates your character editor's live preview

PR #1203, opened 2024-11-10
Add Missing `DOTNET_ROOT` Environment Variable For Nix Devs

DocNITE:
# Description

Added missing environment variable for nix-shell. I don't like everytime write `export DOTNET_ROOT="/path/to/folder"`, so it should be done by post script in `shell.nix`

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -51,5 +51,6 @@ in pkgs.mkShell {
     export ROBUST_SOUNDFONT_OVERRIDE=${pkgs.soundfont-fluid}/share/soundfonts/FluidR3_GM2-2.sf2
     export XDG_DATA_DIRS=$GSETTINGS_SCHEMAS_PATH
     export LD_LIBRARY_PATH=${pkgs.lib.makeLibraryPath dependencies}
+    export DOTNET_ROOT=${pkgs.dotnetCorePackages.sdk_8_0}
   '';
 }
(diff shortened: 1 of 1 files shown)


stellar-novas:
Why do you need to set the DOTNET_ROOT variable? I haven't needed to before

stellar-novas:
> > Why do you need to set the DOTNET_ROOT variable? I haven't needed to before
> 
> It's needed if you have a non-standard dotnet installation. I have this variable defined in my /etc/profile (linux).

I'm a nix user and I don't need to set it, that's why I'm confused

shell.nix
```diff
@@ -51,5 +51,6 @@ in pkgs.mkShell {
     export ROBUST_SOUNDFONT_OVERRIDE=${pkgs.soundfont-fluid}/share/soundfonts/FluidR3_GM2-2.sf2
     export XDG_DATA_DIRS=$GSETTINGS_SCHEMAS_PATH
     export LD_LIBRARY_PATH=${pkgs.lib.makeLibraryPath dependencies}
+    export DOTNET_ROOT=${pkgs.dotnetCorePackages.sdk_8_0}
```
DEATHB4DEFEAT, line 54:
```suggestion
    export DOTNET_ROOT=${pkgs.dotnetCorePackages.sdk_8_0_1xx}
    export PATH="$PATH:/home/$(whoami)/.dotnet/tools"
```

DEATHB4DEFEAT (changes requested):

DEATHB4DEFEAT:
> I'm a nix user and I don't need to set it, that's why I'm confused

I needed to set it to use dotnet tools to create a migration for #1164

shell.nix
DocNITE: Done!

stellar-novas (approved):

DEATHB4DEFEAT (approved):

PR #1209, opened 2024-11-11
Add Tails to IPC

Suraru:
From floofstation, original PR by Fasana

https://github.com/Fansana/floofstation1/pull/253


Aidenkrz:
This doesn't make sense to me.

Resources/Prototypes/Voice/Voice - Shortcut.lnk
DEATHB4DEFEAT, line 1: what is this?

Suraru:
This adds tails to the markings section for IPCs. So you can basically have a Kerfur

Aidenkrz:
I don't think IPCs should have organic tails, if some cool ones were designed for them I'd think that would work but as it stands I don't like this.

DangerRevolution:
Why violate the core design principle of an autonomous android and have a Kerfur (picture attached); when you could have just the autonomous android (with a good and unique design); or add robotic markings to Felinids (as many people have requested)?

Ichaie:
This has already been done in Estação Pirata, and from the experience and feedback of some players, it is not such a good idea, if you are going to add "organic" tails it will not make much sense, I believe that adding tails for example to vulps and felinids will end up taking away a unique characteristic of a playable species, but it might be cool to add original models for IPC

Aidenkrz:
> Why violate the core design principle of an autonomous android and have a Kerfur (picture attached); when you could have just the autonomous android (with a good and unique design); or add robotic markings to Felinids (as many people have requested)?
> 
> 

kerfur

DEATHB4DEFEAT:
I think this is a good idea, but organic tails on them are weird. You can add tails meant for robots if you want.

PR #1233, opened 2024-11-16
Fix Loadout Subcategories Not Being Found

DEATHB4DEFEAT:


VMSolidus (approved):
I've already tested it, this works.
image

FoxxoTrystan (approved):

PR #1220, opened 2024-11-14
Mass Engine Update

sleepyyapril:
# Description

An attempt to do all engine updates in one go. Not focusing on content associated, only what's needed for it to run and any bug fixes.

---

# TODO

- [x] Fix bug where unbuckling resets you to lying down.
- [x] Fix bug where you can no longer get up after lying down.
- [x] See what else I broke.
---

PR #1245, opened 2024-11-18
Hotfix Stamina System

sleepyyapril:
# Description

Goal of this PR is to fix the stamina system, as was caused by #1220.

---

# Changelog

:cl:
- fix: Fixed stamina system


VMSolidus (approved):
Yep, that would do it.

Aidenkrz (approved):

PR #1248, opened 2024-11-18
Glimmer Wisps Mindbreak Their Victims

VMSolidus:
# Description

Small oversight that Glimmer Wisps are generally described as drinking their victim's soul and obliterating their entire personhood. Which is fundamentally just mindbreaking said person. So this PR makes it so that Glimmer Wisps killing a psion also inflicts the Mindbroken condition on them. 

# Changelog

:cl:
- add: Glimmer Wisps now completely obliterate their victim's Personhood, inflicting the Mindbroken condition on them.


VMSolidus:
Test fails aren't real, how the fuck did I get 4 Heisentest fails at once!?

Peptide90 (approved):

PR #1293, opened 2024-11-29
Graytide Virus Event

VMSolidus:
# Description

This is a simple event remake of the classic Graytide Virus event from SS13. A virus infects "Some of the airlocks" on the station, but more specifically it infects the security airlocks, causing them to emag themselves after a short delay. As the name of the airlock virus implies, this lets the prisoners out, and the greytiders in, until an engineer can get to sec and repair the airlocks. It can be a rare and somewhat very literal "Get Outa Jail/Second Chance" for traitors who are caught and permabrigged earlier in the round. 

# Changelog

:cl:
- add: Added the GR4YT1D3 Virus event.


Content.Server/StationEvents/Components/AirlockVirusRuleComponent.cs
```diff
@@ -0,0 +1,19 @@
+using Content.Server.StationEvents.Events;
+
+namespace Content.Server.StationEvents.Components;
+
+[RegisterComponent]
+public sealed partial class AirlockVirusRuleComponent : Component
+{
+    /// <summary>
+    ///     The maximum amount of time in seconds before each infected door is self-emagged.
```
sleepyyapril, line 9: Same description as the field below?

sleepyyapril (changes requested):

sleepyyapril (approved):

Remuchi (approved):

Content.Server/StationEvents/Events/AirlockVirusRule.cs
```diff
@@ -0,0 +1,61 @@
...
+            var parent = Transform(airlockUid).GridUid;
+            if (parent is null
+                || !stationGrids.Contains(parent!.Value))
+                continue;
+
+            airlocks.Add(airlockUid);
+        }
+        foreach (var target in airlocks)
+            Timer.Spawn(TimeSpan.FromSeconds(_random.NextDouble(component.MinimumTimeToEmag, component.MaximumTimeToEmag)), () =>
+                _emag.DoEmagEffect(uid, target));
```
Remuchi, line 51: you can probably remove the whole `airlocks` list and do it inside the `while` block

VMSolidus: I'm fatally allergic to while loops lmao. I'm horribly foreachpilled.

Remuchi: you're iterating over the airlocks twice basically. First time you're gathering them in `while` querry, next time you're applying effects to them. This is simply redundant

VMSolidus: Ahhh. Actually I see how you're right. gimme a few minutes to drop the PR I'm about to drop and I'll do this right away.

sleepyyapril (approved):

Content.Server/StationEvents/Events/AirlockVirusRule.cs
VMSolidus: Got it! There is now no longer a second loop. If an airlock passes the check, it targets that airlock with the emag timer.

Aidenkrz:
https://github.com/space-wizards/space-station-14/pull/33547

PR #1301, opened 2024-12-01
♂️ Add Male Arachne

angelofallars:
# Description

Adds the ability for Arachne to be male.

Also in this I added chest/arm Arachnid markings for Arachne to use. I enabled footstep sounds for Arachne since it doesn't seem that they were intended to have silent footsteps.

## Changelog

:cl: Skubman
- add: Arachne can now be Male instead of only being Female. They can now also wear Facial Hair.
- add: Arachne can use Arachnid chest and arm markings.
- fix: Arachne will now have footstep sounds instead of being quiet when moving.


sleepyyapril (approved):
Mostly repetition, this loooks fine.

VMSolidus (approved):

PR #1307, opened 2024-12-03
The Throwing Update

angelofallars:
# Description

Turns a plethora of items into throwing weapons that deal damage when thrown. Throwing weapons cost stamina to throw.

## Technical/Balance Details

To make a melee weapon also a throwing weapon, just add `- type: DamageOtherOnHit`, and it will automatically inherit the damage from a light melee attack and the melee sound effect as the thrown hit sound effect. You can set a custom damage value with the `damage` field (necessary when the item is not a `MeleeWeapon`) and stamina cost with `staminaCost`.

To make the throwing weapon embed and deal damage over time when embedded, add `- type: EmbeddableProjectile` and `- type: EmbedPassiveDamage`. By default, the embed damage per second is 5% of the throwing damage, but it can be modified on `EmbedPassiveDamage` with `throwingDamageMultiplier`.

The default stamina cost for throwing is 3.5 stamina. The baseline cost for almost all DoT embeddables is 5 stamina, because of the extra damage the DoT brings.

When a thrown item hits a target with body parts, it will randomly select a body part and only deal throwing damage to that body part. It will also embed to the same body part and only deal passive embed damage to it.

## TODO

The unchecked checkmarks are best addressed in another PR but they will stay here for now.

<details><summary>Show Todo</summary>

- [ ] Deal with prediction issue on embeddable projectile removal
  - [ ] This happens even before this PR so not really a big issue, maybe in a separate PR
- [x] Add embeddable damage numbers to embeddables
- [x] Fix throwing angle for surgery tools after the surgical tools sprite update
- [ ] Try to make the throw knockback function as if it hit a wall 
- [x] Esword/desword/e-dagger toggle embed damage
- [x] Don't start passive embed damage if EmbedPassiveDamageComponent has no damage 
- [x] Make DamageOtherOnHit.Damage not nullable
- [x] Throwing damage only to a specific body part

### Traits

- [ ] **Enraged Throw** (Oni)
  - [ ] Oni/trait damage bonus applied to throwing weapon too
  - [ ] Can throw carried bodies, which will do a MassContest between the thrown body and the hit body to determine blunt damage, and stun duration for each party
  - [ ] 15% resistance to thrown/embed damage
    - This helps when their enemy uses the items they throw against them.

- [ ] **Sharpthrower** (Human)
  - [ ] 10% more Brute thrown damage
  - [ ] 50% chance of throw hitting targetted body part
  - [ ] 40% throwing stamina cost reduction
  - [ ] 15% resistance to thrown/embed damage

### Embeds

- [x] Adjust embed damage per second to be like /tg/ (in /tg/ spear has around ~1.2 embed DPS, adjust for ~45% embed chance since we're not implementing embed chance and its 0.54) 
- [ ] Merge EmbeddableProjectileComponent and EmbeddablePassiveDamageComponent
- [ ] Split SharedProjectileSystem into EmbeddableProjectileSystem
- [x] Embed to a specific body part and deal damage only to that part, for now can randomly select body parts on embed
- [ ] ~~Normal passive damage becomes x0.2 when lying down~~
- [ ] Increased damage when moving, more bonus damage for running (Jostle DPS on /tg/ is 0.2 running and 0.1 when walking/crawling)
- [x] All embeddables have a fall out time (30 or 45 secs)
- [ ] - [x] On damage examine, can see that an object is embeddable "It can embed on a target if thrown."
- [ ] Negative moodlet for attached harmful embeddables
- [ ] On health examine target with embeds, can see embedded objects "He has a spear embedded in his left arm."
- [x] On examine item that is embedded, can see to which body part the item is embedded "The spear is embedded on Urist McHands's left arm."
- [ ] An embeddable removed outside of surgery deals a lot of damage (x2 thrown damage)
   - [ ] Lying down prevents natural falling out and thus the damage with non-surgical removal 
- [ ] Surgical procedure on a body part to remove all embeds on it, using hemostat for removal
- [x] Allow anyone to remove embedded cultist weapons even if they're not a cultist

</details>

## Media

**Throwing Toolbox Tools**

https://github.com/user-attachments/assets/4e20568f-adf0-4be8-ac38-fc6b21fed03c

**Examine**

image

**Examine After Embedding**
image

## Changelog

:cl: Skubman
- add: The Throwing Update is here. You can throw most melee weapons at the cost of stamina to deal damage from afar.
- add: Dozens of throwable weapons, mainly sharp weapons will now embed on throw and deal damage every second until they're manually removed or naturally fall off after some time.
- add: Examining the damage values of an item now shows its throwing damage, throwing stamina cost, whether or not it embeds on a throw, and if the embed deals damage over time. 
- add: Examining an embedded item now shows what body part it's embedded in.
- tweak: The traits High Adrenaline, Adrenal Dysfunction, Masochism and Low Pain Tolerance now affect throwing attacks just like melee attacks.
- tweak: The default time to remove embedded items has been increased from 3 to 5 seconds, and both the remover and the target with the embedded item need to stand still during the removal.
- tweak: The time to pry up a floor tile with a crowbar and other tools has been decreased from 1 second to 0.5 seconds. The throwing damage of floor tiles has been increased. Go figure.
- fix: Attempting to throw a Blood Cultist item without being a cultist will stun you and drop the item you're holding properly.


BlueHNT:
blunt weapons bounce off of people?

angelofallars:
> blunt weapons bounce off of people?

Yes they do

https://github.com/user-attachments/assets/4e20568f-adf0-4be8-ac38-fc6b21fed03c

VMSolidus:
LET'S FUCKING GOOOOOOOO

VMSolidus:
https://github.com/Simple-Station/Einstein-Engines/issues/467

BlueHNT:
I love everything about this

VMSolidus:
image

This entire method is not allowed on ComponentStartup, because modifying the datafields of a component during component startup will cause corruption of map.yml files(It will *save* the component modifications to the map if you do this on a live server). You can instead do this through MapInitEvent, which is the semifinal initialization step, and occurs after the maps are saved.

angelofallars:
> image
> 
> This entire method is not allowed on ComponentStartup, because modifying the datafields of a component during component startup will cause corruption of map.yml files(It will _save_ the component modifications to the map if you do this on a live server). You can instead do this through MapInitEvent, which is the semifinal initialization step, and occurs after the maps are saved.

Gotcha will try that out

Content.Server/Flash/FlashSystem.cs
```diff
@@ -60,7 +62,7 @@ private void OnFlashMeleeHit(EntityUid uid, FlashComponent comp, MeleeHitEvent a
         {
             if (!args.IsHit ||
                 !args.HitEntities.Any() ||
-                !UseFlash(uid, comp, args.User))
+                !UseFlash(uid, comp))
             {
                 return;
```
sleepyyapril, line 67: nitpick; remove brackets around 1-line returns when you run into them

Content.Server/Projectiles/ProjectileSystem.cs
```diff
@@ -77,4 +80,22 @@ private void OnStartCollide(EntityUid uid, ProjectileComponent component, ref St
...
+    private void OnDamageExamine(EntityUid uid, EmbeddableProjectileComponent component, ref DamageExamineEvent args)
+    {
+        if (!component.EmbedOnThrow)
+            return;
+
+        if (!args.Message.IsEmpty)
+            args.Message.PushNewline();
+
+        var isHarmful = TryComp<EmbedPassiveDamageComponent>(uid, out var passiveDamage) && passiveDamage.Damage.Any();
+
```
sleepyyapril, line 93: nitpick: you could remove the space here considering the code is similar and used right below.

Content.Shared/Damage/Components/DamageOtherOnHitImmuneComponent.cs
```diff
@@ -0,0 +1,10 @@
+namespace Content.Shared.Damage.Components
+{
+    /// <summary>
+    /// This is used for entities that are immune to damage from DamageOtherOnHit.
+    /// </summary>
+    [RegisterComponent]
+    public sealed partial class DamageOtherOnHitImmuneComponent : Component
```
sleepyyapril, line 7: nitpick: put { } on the same line as the class definition

Content.Shared/Damage/Systems/SharedDamageOtherOnHitSystem.cs
```diff
@@ -0,0 +1,203 @@
...
+                if (HasComp<MobStateComponent>(args.Target))
+                    _adminLogger.Add(LogType.ThrowHit, $"{ToPrettyString(args.Target):target} received {modifiedDamage.GetTotal():damage} damage from collision");
+
+                _meleeSound.PlayHitSound(args.Target, null, SharedMeleeWeaponSystem.GetHighestDamageSound(modifiedDamage, _protoManager), null,
+                    component.SoundHit, component.SoundNoDamage);
+            }
+
+            if (modifiedDamage is { Empty: false })
+            {
+                _color.RaiseEffect(Color.Red, new List<EntityUid>() { args.Target }, Filter.Pvs(args.Target, entityManager: EntityManager));
```
sleepyyapril, line 109: nitpick: remove squiggles

sleepyyapril (approved):

VMSolidus (approved):

sleepyyapril (approved):

PR #1322, opened 2024-12-07
CVar CharacterRequirement

sleepyyapril:
# Description

This takes in one CVar and one required value and simply checks if the CVar's value as a string is equal to the required value. I could make it use a list of them, but I really didn't think it was needed considering CharacterAndLogic.

I also cleaned up CharacterRequirements.Profile.cs a bit.


Content.Shared/Customization/Systems/CharacterRequirements.Misc.cs
```diff
@@ -0,0 +1,55 @@
...
+using Content.Shared.Roles;
+using JetBrains.Annotations;
+using Robust.Shared.Configuration;
+using Robust.Shared.Prototypes;
+using Robust.Shared.Serialization;
+
+namespace Content.Shared.Customization.Systems;
+
+/// <summary>
+///     Requires the profile to have one of the specified loadouts
```
FoxxoTrystan, line 12: not sure if this is the right summary

sleepyyapril: good catch

Content.Shared/Customization/Systems/CharacterRequirements.Misc.cs
```diff
@@ -0,0 +1,55 @@
...
+
+namespace Content.Shared.Customization.Systems;
+
+/// <summary>
+///     Requires the server to have a specific CVar value. 
+/// </summary>
+[UsedImplicitly, Serializable, NetSerializable,]
+public sealed partial class CVarRequirement : CharacterRequirement
+{
+    [DataField("cvar", required: true)]
```
VMSolidus, line 17:
```suggestion
    [DataField(required: true)]
```

Content.Shared/Customization/Systems/CharacterRequirements.Misc.cs
```diff
@@ -0,0 +1,55 @@
...
+    public string CVar;
+
+    [DataField("requiredValue", required: true)]
```
VMSolidus, line 20:
```suggestion
    [DataField(required: true)]
```

VMSolidus (changes requested):

Resources/Locale/en-US/customization/character-requirements.ftl
```diff
@@ -140,3 +140,11 @@ character-whitelist-requirement = You must{$inverted ->
     [true]{" "}not
     *[other]{""}
 } be whitelisted
+
+## Cvar
```
DEATHB4DEFEAT, line 144:
```suggestion
## CVar
```

DEATHB4DEFEAT (approved):

PR #1339, opened 2024-12-12
Enhances Roundstart Antagonist Selection to Prevent Roundstart Fail and Allocate All Antags

rbertoche:
We might never see the end of new antags being added using MakeAntag so better just cut it off by the root and mitigate the issue by preventing it from causing the round not to start.

# Description

Since [this issue](https://discord.com/channels/1218698320155906090/1218701887990140999/1314367545226166333) still persisted to manifest even after an attempt to fix it, this PR simply prevents the rounds from not starting due to the error by changing the exception into a log error.

Catching and treating could be done instead, but I don't see many benefits from that at all, so an error it is. 

Also important, after asking a player in ahelp about what settings they had, I concluded it gets triggered when a player has Thief antagonist enabled, and have jobs set up in such a way that they remain in lobby even when they ready up. So they seem "elligible" but they have no entity at roundstart due to their setting to not get spawned at roundstart as assistant and wait in lobby when you don't have the jobs you had selected.

I tried to figure a way to still add the same antagonist it failed to assign in [ChooseAntags](https://github.com/Simple-Station/Einstein-Engines/blob/0df6205a88ffcf995e242e9225d6bf8085135b73/Content.Server/Antag/AntagSelectionSystem.cs#L202), and considered decrementing `i` inside that loop, but then I didn't due to how decrementing `i` would make the loop prone to stalling. So we lose one of the antagonist slots, but get the round to start, best effort.

Cheers.
-----

Edit: Please read [my last comments](https://github.com/Simple-Station/Einstein-Engines/pull/1339#issuecomment-2558363980).

---

# TODO

---

<p>

Example Media Embed

</p>
</details>

---

# Changelog

:cl:
- tweak: Enhances roundstart antagonist selection making sure all slots are allocated
- fix: Prevents Thief or others antags from causing the round to fail to start


rbertoche:
I tried to figure a way to still add the same antagonist it failted to assign in ChooseAntags, and considered decrementing i inside that loop, but then I didn't due to how decrementing i would make the loop prone to stalling. So we lose one antagonist, but get the round to start, best effort.

Remuchi:
Current implementation can lead to no antags being selected in the round at all. IMO if this error happens, the antag selection should reselect player for that role

rbertoche:
Agreed, but given the reason why it fails to assign I didn't figure a way to retry to assign the role without risking freezing on that loop on the case every candidate is not going to pass the checks.
I would do a change in that direction otherwise

rbertoche:
Well.
I can do a silly hack and like retry 5 times. That would retry, and would not risk a forever stalling loop. Is that a better change?

rbertoche:
Set a constant to determine the max number of attempts. Retry until that number or until the required slots are assigned. It wouldn't be too bad. I can do that later tonight.

sleepyyapril (changes requested):
Make it return a bool. If it fails, try try again up to a constant number of tries.

rbertoche:
Would changing the signature of MakeAntag for this be worth it? It would make EE antag code less compatabol.
A new method on the other hand would have less potential of requiring conflicts.
For one I could do a new method, keep the exception and intercept it... But it would have the downside of not working properly for code which just calls MakeAntag. I am not even sure other code calls it... Maybe I am worrying for nothing.
But right now I would think making a retry count inside ChooseAntags would be the less impactful change which still would fix not picking enough antags.
But if you insist I will just make it return bool sure.

sleepyyapril:
> Would changing the signature of MakeAntag for this be worth it? It would make EE antag code less compatabol. A new method on the other hand would have less potential of requiring conflicts. For one I could do a new method, keep the exception and intercept it... But it would have the downside of not working properly for code which just calls MakeAntag. I am not even sure other code calls it... Maybe I am worrying for nothing. But right now I would think making a retry count inside ChooseAntags would be the less impactful change which still would fix not picking enough antags. But if you insist I will just make it return bool sure.

Mm, true, should be fine as is.

rbertoche:
So. I ended up changing MakeAntag a lot while preserving APIs. But now I need to properly test this, since it's not tiny anymore.
I think previous code was not great as a consequence of the lack of features on AntagSelectionPlayerPool - or maybe it shouldn't have become a named type at all and having to duplicate Random and List methods is a code smell? Anyway
The added methods make it way easier to write proper code to get N antags, and to filter the pool in a sane way.

rbertoche:
There's only one thing which I'm not sure about these last changes. Calling MakeAntag(ent, null, def) to fill up unused slots makes sense, but it may have unintended consequences.
What is better? Just leave some antags missing if count does not get to 0, or calling MakeAntag(ent, null, def)?

rbertoche:
Instead of writing a quick hack I tried to make MakeAntag better in a non-hacky way.

rbertoche:
Tested it briefly with a single connection. It did add Thief at roundstart with no issues and didn't break when there weren't any thief available at roundstart either.

rbertoche:
https://github.com/Simple-Station/Einstein-Engines/actions/runs/12452770702/job/34762138592#step:10:405

This looks relevant but I am not sure how to fix it. It's the Log.Error in the changes.

rbertoche:
I think this error is supposed to happen in the tests. This new ChooseAntags will detect the failure and try to assign the roles, so I could change it from an error to a info or at least a warning. But since I already copied the state of that code region from wizden I don't think it's worth it to change it to reduce the log level.

Feel free to fix the test as you think is appropriate and merge if you want, if it's convenient

sleepyyapril:
> https://github.com/Simple-Station/Einstein-Engines/actions/runs/12453149394/job/34762931010#step:10:362
> 
> This looks relevant but I am not sure how to fix it. It's the Log.Error in the changes.

Could check if the player is valid, like IsTerminatingOrDeleted, and if not choose another.

rbertoche:
Right, after you say it is seems so obvious :P
I assumed there weren't enough candidates in the test, I'm gonna try that

Content.Server/Antag/AntagSelectionPlayerPool.cs
```diff
@@ -23,5 +24,46 @@ public bool TryPickAndTake(IRobustRandom random, [NotNullWhen(true)] out ICommon
...
+    // EE
+    public bool TryGetItems(IRobustRandom random,
+                            [NotNullWhen(true)] out ICommonSession[]? sessions,
+                            int count,
+                            bool allowDuplicates = true)
+    {
+        DebugTools.Assert(count > 0, $"The count {nameof(count)} of requested sessions must be greater than zero!");
+
+        sessions = null;
+        List<ICommonSession> session_list = [];
```
Timfa2112, line 36:
```suggestion
        var session_list = new List<ICommonSession>();
```

Content.Server/Antag/AntagSelectionSystem.cs
```diff
@@ -221,20 +221,68 @@ public void ChooseAntags(Entity<AntagSelectionComponent> ent,
...
+            // Tries multiple times to assign antags.
+            // When any number of assignments fail, next iteration
+            // gets new items to replace those.
+            // Already selected or failed sessions are avoided.
+            // It retries until it ends with no failures or up
+            // to maxRetries attempts.
+
+            const int maxRetries = 4;
+            var retry = 0;
+            List<ICommonSession> failed = [];
```
Timfa2112, line 243:
```suggestion
            var failed = new List<ICommonSession>();
```

Timfa2112 (changes requested):
Two style comments, the rest seems OK
Have you verified that this works? A video of it doing what it needs to do (if that's possible for this PR) would be nice

Timfa2112 (approved):

PR #1340, opened 2024-12-12
Port Cosmatic Drift Arrivals

VMSolidus:
# Description

Done after a lengthy discussion with @OldDanceJacket about mapping issues, namely the existence of "Walk-In Closets" on maps that take up an inordinate amount of space that would otherwise be usable for other purposes. Seriously, Saltern has a 4x6 room that contains >200 medical uniforms, and is on a map intended only for 15 people(And maximum of 6 medbay personnel). Every map in rotation is like this.

Ostensibly, Loadouts were created to provide a solution to this problem, by making it so that players choose their DRIP before spawning in, so that mappers don't need to put in a whole ass room dedicated to medical uniforms. But Mappers ended up keeping the "Walk-in Closets" even after Loadouts were introduced to serve as a "Sanity check" in case of events such as people not choosing their loadouts, or getting hired during the round by the HoP to work in the medbay, etc. To solve this, me and ODJ agreed upon adopting Cosmatic Drift's solution to this, by making it so that the Arrivals Terminal serves as the "Walk-in Closet Hotel" that every map can just piggyback off of. Nobody needs to make the stupid closet rooms, on ANY map, because every map is fully guaranteed to have every closet the players could ever need in arrivals. 

# Media
<p>

Dock
image

Dorms(Spawn points)
image

Medbay: 
image

Security
image

Captain/HOP
image

Cargo:
image

Epistemics:
image

</p>
</details>

# Changelog

:cl:
- add: Ported Cosmatic Drift's version of the Arrivals Terminal. Terminal now includes every version of department clothing and material lockers that players could ever need, allowing them to try on their chosen job equipment and drip before departing for the station.


sleepyyapril:
Requesting media for this one.

VMSolidus:
> Requesting media for this one.

Actually more importantly I'm trying to figure out why rounds don't start now. And I assume it's because this map is missing some necessary component.

sleepyyapril (approved):
At the very least, this is an improvement to current arrivals.

VMSolidus:
By Peptide's suggestion, I replaced every single computer in the entire Terminal with a BaseComputer, which does nothing and is completely ornamental. I also replaced all the dressers in each head of staff room with a dresser containing that head of staff's uniforms.

sleepyyapril:
I do have a few concerns over low-pop activity in this area, but I think overall the fact that most people will be on-station will encourage them to go anyway, so long as the spawn is near the arrivals area. Having a bigger place to chat while you wait for the shuttle is no issue.

Remuchi (approved):

PR #1346, opened 2024-12-14
Fix Interact Verb

sleepyyapril:
Finally fixed it.

Interact verbs weren't appearing because this is required in BaseMob.


VMSolidus (approved):

Remuchi (approved):

Remuchi:
I believe you

PR #1348, opened 2024-12-14
Pointy Ears For Harpies And Arachne

VMSolidus:
# Description

Like half the art references I have for Harpies have them with elf-like pointed ears, another half have them with their feather tufts. We have pointy ears markings, so this PR makes it so that Harpies can have them. Actually it also gives them to Arachne, for similar reasons(It's an oddly common thing on Arachne/Lamia art depictions). 

# Changelog

:cl:
- add: Added a variety of Pointy Ears markings for Harpy and Arachne characters.

### DIFF
--- a/Resources/Prototypes/Entities/Mobs/Customization/Markings/pointy_ears.yml
+++ b/Resources/Prototypes/Entities/Mobs/Customization/Markings/pointy_ears.yml
@@ -3,7 +3,7 @@
   bodyPart: HeadSide
   markingCategory: HeadSide
   forcedColoring: true
-  speciesRestriction: [Oni, Harpy]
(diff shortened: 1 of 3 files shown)


sleepyyapril (approved):

angelofallars (changes requested):
Need to make sure Harpy and Arachne have at least 3 Head (Side) markings so they can have ears and two earrings at the same time.

VMSolidus:
> Need to make sure Harpy and Arachne have at least 3 Head (Side) markings so they can have ears and two earrings at the same time.

Done

angelofallars (approved):

Remuchi (approved):

PR #1355, opened 2024-12-16
Fix Redshirt & Brittle Bone Traits

VMSolidus:
These were supposed to be negative traits(They give you points), but I accidentally made them positive traits(They cost points).


sleepyyapril (approved):
silly!

DEATHB4DEFEAT (approved):

PR #1356, opened 2024-12-17
Example Hub Ad List

VMSolidus:
# Description

this PR extends the list of default advertiser hubs to include every existing hub except for SSMV(Which is mutually exclusive with the others). If there are more hubs I have missed, please let me know. This will allow our servers to appear on as many hubs as possible.


Erisfiregamer1 (approved):
SPEED IT UP

DEATHB4DEFEAT (approved):

Evgencheg (approved):
WE NEED THIS!

sleepyyapril (approved):

Issue #1379, opened 2024-12-28
Europa requested fixes


BlueHNT:
# Description

So... [Europa](https://github.com/Simple-Station/Einstein-Engines/pull/1152) is out and it's pretty neat, but as engineer it's a nightmare.

Gonna list 'em out and it's gonna be a long one.

1. The SM Issues
So beginning with the whole issue that the passive cooling is useless is one thing as the planet has roughly 80-90°C and that won't work for the passive cooling AT ALL. I can manually fix it by closing it off with a buncha heaters, but normal people won't even notice it.
![image](https://github.com/user-attachments/assets/c9c89347-05cf-4031-a01a-45adaa359ef7)

On top of that, all engineering equipment is hidden in the SM room which is a bloody nightmare. You gotta pass by and lose your birthrights to even get your hardsuit.
![image](https://github.com/user-attachments/assets/eeac31b1-b6cf-4397-8e24-676892312f31)

Not to mention the nonsensical lockup of random equipment that doesn't even matter and a few tesla coils.
![image](https://github.com/user-attachments/assets/a6f5ac23-2dd3-49a3-b194-b8ab068621dc)

2. The Atmos and their issues
Oh boy the atmos, first of all it comes prepared already in it's tight packaged corner while at the same time missing all the features a regular atmos would need for this size of map. It even comes pre-made in a mega simple way, which means 90% of Atmos techs will just run some air and fuck off.
![image](https://github.com/user-attachments/assets/7df7e386-5d33-4809-aba5-cdee693587f5)

Next up the miner for Plasma and N2O is outright missing! Meaning you can't make tritium for engine, build TEG or make any normal SM mix.
![image](https://github.com/user-attachments/assets/1a07c344-d396-4ca2-b12c-1bafdb0889d3)

The little place for dedicated SM use is cute however.
![image](https://github.com/user-attachments/assets/afae6949-9014-462a-8880-aa5b0414340c)

Oh and roundstart the mixer chamber is missing Blast Doors to outside to vent your used mixes, while also missing the button for the already existing Blastdoor which however features a gracious tinyfan below it. 
![image](https://github.com/user-attachments/assets/aa29c12c-5d57-49f7-8341-28b0dc582aa9)

Additionally the map is also missing atmos techs hardsuits, meaning they can't fight fire even if they wanted with the non-pressure resistant fire suits they got.
[...]

PR #1380, opened 2024-12-28
Remove Europa From Map Pool

sleepyyapril:
# Description

There are a _lot_ of issues surrounding Europa, which include, but the big two are:
- How many complaints I'm receiving about it in general.
- Engineering issues, including a lack of an alternative beyond an SM (supposedly)
- Lacking a cryosleep area.
- Distro air and waste are connected together.
- More here https://github.com/Simple-Station/Einstein-Engines/issues/1379#issuecomment-2564303908

Is there a way we can add people to review maps specifically? I'm not a mapper, but the last few maps that have gotten through have been _bad_ at the start.

---

# Changelog

:cl:
- remove: *Temporarily* removed Europa from map pool, pending some refurbishing.

### DIFF
--- a/Resources/Prototypes/Maps/Pools/default.yml
+++ b/Resources/Prototypes/Maps/Pools/default.yml
@@ -16,4 +16,4 @@
   - TheHive
   - Gax
   - Rad
-  - Europa
+#  - Europa


VMSolidus (approved):

Remuchi (approved):
Fair

PR #1394, opened 2025-01-02
More Psionic Admin Commands

VMSolidus:
# Description

This PR adds three new admin commands that make use of the new functionality available to Psionics V3. These commands are:

- addrandompsionicpower
This takes a Uid and throws it at the random psi power generation code. Which gives it a random psi power that is available in its list of rollable powers.

- removepsionicpower
This takes a Uid and a PsionicPower prototype ID, and attempts to remove that power from said entity. It does nothing if the entity doesn't have that power.

- removeallpsionicpowers
As above, but instead of needing to declare a power prototype, it just removes ALL powers that the entity has, if any.

# Changelog

:cl:
- add: Added new admin commands AddRandomPsionicPower, RemovePsionicPower, and RemoveAllPsionicPowers.

### DIFF
--- a/Content.Server/Psionics/PsionicsCommands.cs
+++ b/Content.Server/Psionics/PsionicsCommands.cs
@@ -105,18 +105,27 @@ public sealed class RemovePsionicPowerCommand : IConsoleCommand
 
         if (!EntityUid.TryParse(args[0], out var uid))
         {
-            shell.WriteError(Loc.GetString("addpsionicpower-args-one-error"));
+            shell.WriteError(Loc.GetString("removepsionicpower-args-one-error"));
             return;
         }
 
         if (!protoMan.TryIndex<PsionicPowerPrototype>(args[1], out var powerProto))
         {
-            shell.WriteError(Loc.GetString("addpsionicpower-args-two-error"));
+            shell.WriteError(Loc.GetString("removepsionicpower-args-two-error"));
             return;
         }
 
         if (!entMan.TryGetComponent<PsionicComponent>(uid, out var psionicComponent))
+        {
+            shell.WriteError(Loc.GetString("removepsionicpower-not-psionic-error"));
             return;
+        }
+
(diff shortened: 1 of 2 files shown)


sleepyyapril (approved):

Content.Server/Psionics/PsionicsCommands.cs
```diff
@@ -63,3 +63,97 @@ public async void Execute(IConsoleShell shell, string argStr, string[] args)
...
+    public async void Execute(IConsoleShell shell, string argStr, string[] args)
+    {
+        var entMan = IoCManager.Resolve<IEntityManager>();
+        var psionicPowers = entMan.System<PsionicAbilitiesSystem>();
+
+        if (!EntityUid.TryParse(args[0], out var uid))
+        {
+            shell.WriteError(Loc.GetString("removeallpsionicpowers-args-one-error"));
+            return;
+        }
```
sleepyyapril, line 149: autoparse when...

sleepyyapril (approved):

Resources/Locale/en-US/psionics/psionic-commands.ftl
```diff
@@ -11,3 +11,19 @@ command-addpsionicpower-description = Initialize an entity as Psionic with a giv
 command-addpsionicpower-help = Argument 1 must be an EntityUid, and argument 2 must be a string matching the PrototypeId of a PsionicPower.
 addpsionicpower-args-one-error = Argument 1 must be an EntityUid
 addpsionicpower-args-two-error = Argument 2 must match the PrototypeId of a PsionicPower
+
+command-addrandompsionicpower-description = Initialize an entity as Psionic with a a random PowerPrototype that is available for that entity to roll.
```
DEATHB4DEFEAT, line 15:
```suggestion
command-addrandompsionicpower-description = Initialize an entity as Psionic with a random PowerPrototype that is available for that entity to roll.
```

DEATHB4DEFEAT (approved):

Issue #1151, opened 2024-10-27
ClothingAddFaction Doesn't work well.


Peptide90:
# Description

We found this on N14 when adding the component to IDs to allow players to change factions, however the upstream `ClothingUniformMailCarrier` also uses the component to add Mailman faction.

The issue for us is we use turrets which are hostile to one faction, but friendly to another, and adding the friendly faction doesn't overwrite the existing faction, so people still get shot.

This change or addition would mean you could sneak onto a syndicate ship if you manage to get a syndicate hardsuit or uniform for example, as it will be the same code issue.

# Media

![Example Screenshot](https://example.com/thisimageisntreal.png)

sapphirescript:
This should be solved with #1407 as it allows you to do fancier functions with factions and clothing, but you'd have to try it out. @Peptide90

PR #1407, opened 2025-01-02
Generic Clothing Equip Functions

VMSolidus:
# Description

This PR was originally going to be called "Psionic Refactor V3 Part 2, Items Of Power", but I began getting shakes and started vomiting when I saw the list of Components that make items do things when equipped, and I had a conniption about how much code there is constantly being repeated. So instead of this PR adding Items Of Power, I added a universal modular generic system for making clothing items do things when equipped and unequipped. Which hooks into the library of TraitFunctions that we've previously created. I also added a few "Inverse" versions of trait functions that can be used by these new clothing functions.

### DIFF
--- a/Content.Client/Clothing/ClientClothingSystem.cs
+++ b/Content.Client/Clothing/ClientClothingSystem.cs
@@ -130,8 +130,8 @@ public sealed class ClientClothingSystem : ClothingSystem
 
         RSI? rsi = null;
 
-        if (clothing.RsiPath != null)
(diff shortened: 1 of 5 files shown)


sleepyyapril (approved):

sleepyyapril:
phone review but looks fine

sapphirescript (commented):
This should also solve https://github.com/Simple-Station/Einstein-Engines/issues/1151, as TraitSystem.Functions has methods to modify factions and should accomplish what they're looking for

DEATHB4DEFEAT (approved):

PR #1418, opened 2025-01-03
Make Crit/Death Traits Mutually Exclusive

sapphirescript:
# Description

Red Shirt and BBD didn't follow the pattern of disallowing their opposite traits from being equipped (I.e. tenacity v glass jaw, will to live v will to die, amplification v dampening traits).

---

:cl:
- tweak: You can no longer simultaneously equip Red Shirt and Brittle Bones Disease and their positive counterparts, Will To Live and Tenacity.


VMSolidus (approved):
Technically the traits originally had the limitation for technical reasons, two traits previously could not write to the same component. They no longer have that limitation. But I suppose now there's a new limitation specifically with these traits in particular, making this necessary until I again refactor MobState to make this not an issue anymore.

sleepyyapril (approved):

PR #1429, opened 2025-01-05
[Port] Add Time Transfer Panel From Goobstation

CerberusWolfie:
# Description
This adds the Time Transfer Panel that Goobstation uses. This panel enables Admins to do Time Transfers without commands and makes their lives easier. This immediately has access to all playtime trackers and allows an Admin to Add Time and/or Set Time (overwriting the previous time) for any job role. It also enables extra strings since I ported from Wizden upstream as well, which can be seen in the commit history.

A thing of note when testing this and using it in a server: the times **DO NOT UPDATE** for Client until your **RELOG**. This is how it works on Goobstation as well. The time adding process is faster and will be easier tool.

This is tied to the AdminFlags.Admin so it can be customized on a per-server basis if a custom flag is added.

---

# TODO
- [x] Port Upstream Role Define.
- [x] Port string extras for playtime commands.
- [x] Port Time Transfer Panel
- [x] Remove Overall (it just doesn't work on Goob either).

---

<p>

image
image
image
image
image
image
image
image
image

</p>
</details>

---

# Changelog
:cl:
ADMIN:
- add: Add Time Transfer Panel


sleepyyapril (approved):

Aidenkrz (approved):

CerberusWolfie:
There. I force-pushed upstream merge into my branch so it stops bitching about merge conflicts.

Remuchi (approved):

PR #1430, opened 2025-01-05
Add Different Launch Types to VSCode

CerberusWolfie:
# Description
This adds several different launch types to the VSCode. This allows you to launch Client in Compatibility Mode for the Tools/Release versions as well as launch the Server/Client at the same time in Tools/Release modes and use the VSCode Terminal.

---

# TODO
- [x] Add different Build types in the tasks.json file.
- [x] Add different Launch types in the launch.json file.

---

image

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -4,6 +4,7 @@
     // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
     "version": "0.2.0",
     "configurations": [
+        // CLIENT CONFIGS
         {
             "name": "Client",
             "type": "coreclr",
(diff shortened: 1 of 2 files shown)


VMSolidus (approved):
Can you also add one that runs dotnet test?
It's always irritated the fuck out of me that I can't use breakpoints during the tests when ran locally, which makes rooting out heisenbugs essentially impossible.

CerberusWolfie:
Uh, yeah, give me a bit to grab it.

VMSolidus:
Actually Durk pointed out to me something I didn't know, and that's not necessary at all it seems. So you're good!

PR #1351, opened 2024-12-15
Wizmerge for Station AI

sleepyyapril:
# Description

the adding AI is now up to y'all because i'm not touching loadout code for name datasets, but it shouldn't be too bad from here

PR #1434, opened 2025-01-05
Throwing Mini-Update 1

angelofallars:
# Description

Bug fixes:

- Re-enable damage examine for embeddable passive damage after the Wizmerge for Station AI wiped it (https://github.com/Simple-Station/Einstein-Engines/pull/1351) 
- Fixed a bug where emags did not have embed passive damage (due to `DamageOtherOnHitStartupEvent` only raising when `MeleeWeaponComponent` exists on the entity)

Tweaks:

- Added a stamina cost to spears, 6 stamina, and removed spears dealing damage to itself when thrown
- Floor tiles will now break when thrown 10 times instead of 4 times
- Added the MetalThud sound effect for structures when thrown, and increased the base thrown damage of structures slightly from 8 to 9

Damage examine is now sorted properly. From top to bottom:
- Gun damage
- Melee weapon damage
- Throwing weapon damage
- Embed passive damage

## Media

**Laser rifle sorted damage examine**

image

**Spear sorted damage examine**

image

**Emag stats**

image

## Changelog

:cl: Skubman
- fix: Fixed embeddable damage over time not showing when examining an item's damage values.
- fix: Fixed the emag not dealing passive damage when embedded on a target. 
- tweak: Examining an item's damage values now shows values in a sorted order. From top to bottom: gun damage, melee damage, throwing damage, embedded damage.
- tweak: Spears now have a 6 stamina cost when thrown, and they no longer break when thrown enough times.
- tweak: Floor tiles now break when thrown 10 times instead of 4 times.
- tweak: Closets, lockers and crates now play the proper sound effect when thrown (by Space Wind).


VMSolidus (approved):
Wild that this got disabled accidentally

PR #1471, opened 2025-01-09
Some Surgery Fixs

Spatison:
# Description

When attempting to insert an organ, a check is now performed to ensure a container exists for it. Skeletons can now lie down, and nymphs can now be subject to surgery.
---

# Changelog

:cl: Spatison
- tweak: Nymphs are now operable.
- fix: Skeletons can now lie down.
- fix: It is no longer possible to insert an organ into a body if the required space is unavailable.


angelofallars (changes requested):

VMSolidus (approved):
Out of curiosity, is it possible to kitbash a human head & brain onto a slime who has a brain in their torso(slime core)?

PR #1476, opened 2025-01-10
Senior Physician CPR

VMSolidus:
# Description

Senior Physician wasn't included in the list of jobs that can't take CPR Training, but they already have CPR training, thus allowing players who don't know this to waste 4 points on a trait they get for free.


Remuchi (approved):

BlueHNT:
What about medborgs?

PR #1507, opened 2025-01-12
Cryo Overhaul

neuPanda:
Description.
added 2 intermidiate chems
- SalicylicAcid
- Formaldehyde

added 2 Cryo Chems
-Traumoxadone (Brute Damages)
-Stelloxadone (Tox Damages)

modified existing cryo chems to be more viable yet still balanced improved efficiency of cryo chambers
- Cryox only heals blood loss and air loss damage on a body, now useable after death
- Alox now also heals cold damage
- Doxa now uses Phalanximine not UnstableMutagen works after death twice as effective
- Oppo now uses Formaldehyde not Cognizine

updated cryo guide book to show all cryo chems

modified body metabolism
- the body no longer only metabolizes 2 chems at once but all chems in the system
- restrictive logic still somewhat in tact to prevent poison power gaming max 3 toxins

modified cryopods
- cryopods now split chems and inject them independently rather than a mixed solution
- cryopods require 1/2 the amount of chems to do the same healing

:cl:
- add: new cryo chems
- tweak: old cryo chems
- fix: death
- remove: metabolism restrictions

## Summary by CodeRabbit

- **New Features**
	- Introduced new reagents and chemical recipes, expanding treatment and reaction options.
	- Enhanced cryogenic treatments with improved reagent injection and potency scaling.
- **Refactor**
	- Adjusted substance processing to differentiate between general components and hazardous ones.
- **Documentation**
	- Updated in-game guides and reagent descriptions to reflect new thresholds and treatment parameters.


Remuchi:
> the body no longer only metabolizes 2 chems at once but all chems in the system
I'm really not a fan of this change since it was a feature for races like slimes. They could metabolize up to 5 chems at once

neuPanda:
was porting this up from past implementation before surgery and metabolize need to look at this again in the morning

gluesniffler:
I feel dividing cryo into multiple chems is a bit silly for brute/airloss/etc. Cryoxadone should be the general use chem but with varying efficiency (leaning into slow) based on the temperature of the body like in 13, as cryo is supposed to be dead simple, super easy to make, but somewhat slow if you're not optimizing it, but basically a panacea along with mannitol. I've done this work on a branch for goob but haven't PRd it, feel free to leave it as is and I'll just adjust later.

Also, I quite like these changes to chemistry as it works closer to how ss13 does it (you just metabolize the given rate of all chemicals in your system that are not poisons). Which also lets chemistry do heal-all pills and other funsies. 

Think you could also add the effect from SS13 on salicylic acid too?
image

neuPanda:
> I feel dividing cryo into multiple chems is a bit silly for brute/airloss/etc. Cryoxadone should be the general use chem but with varying efficiency (leaning into slow) based on the temperature of the body like in 13, as cryo is supposed to be dead simple, super easy to make, but somewhat slow if you're not optimizing it. Basically a panacea along with mannitol. I've done this work on a branch for goob but haven't PRd it, feel free to leave it as is and I'll just adjust later.
> 
> Also, I quite like these changes to chemistry as it works closer to how ss13 does it (you just metabolize the given rate of all chemicals in your system that are not poisons). Which also lets chemistry do heal-all pills and other funsies.
> 
> Think you could also add the effect from SS13 on salicylic acid too? image

i could probably add the effects from 13 as well
also you can already test these cryo chems on frontier (altered) floof and the den.  
the reason cryo is being made more rather than less complicated is the goal when i first did this back in frontier probably around April/May time frame, there was no surgery, chems were already super easy, and there was not an acceptable RP way to recover somebody.  and no, i refuse to accept 200lb of ointment wrapped in band aids as an acceptable RP solution 😅

the reason it has taken so long to get here is i had been hoping for more feedback before integrating into upstream.  however feedback has been minimal but the feedback i have gotten has been extremely positive.  not broken but has been enough of a bump to get it put on the radar as a form of treatment but not the 1 and only go-to.  it has at least on the den become more of a "if they are dead use cryo, if they are alive use chems" rather than "always use topicals"

Content.Server/Body/Components/MetabolizerComponent.cs
```diff
@@ -54,14 +54,13 @@ public sealed partial class MetabolizerComponent : Component
...
-
         /// <summary>
-        ///     How many reagents can this metabolizer process at once?
+        ///     How many poisons can this metabolizer process at once?
         ///     Used to nerf 'stacked poisons' where having 5+ different poisons in a syringe, even at low
         ///     quantity, would be muuuuch better than just one poison acting.
         /// </summary>
-        [DataField("maxReagents")]
-        public int MaxReagentsProcessable = 3;
+        [DataField("maxPoisons")]
```
VMSolidus, line 62:
```suggestion
        [DataField("maxReagents")]
```
This is why renaming DataFields is not a good idea.

image

VMSolidus (changes requested):

Content.Server/Body/Components/MetabolizerComponent.cs
Erisfiregamer1: > This is why renaming DataFields is not a good idea.
> 
> image

I think looking at that gave me a stroke

VMSolidus:
@coderabbitai full review
Go ahead and nitpick grammar and punctuation issues too, it's one of my weaknesses anyways.

PR #1521, opened 2025-01-12
Nerf Salvage

VMSolidus:
# Description

I may have been a little *too* generous with Salvage players. This update fixes the salvage guns spawning with a magazine or lethal ammo. There was also an issue that they were perhaps "Too easy to obtain". In actual practice, it turns out that a single "Mining Expedition" earns salvage enough mining points to buy like 4 rifles. Which led to situations where Salvage functionally became station security after they then broke into the armory to print ammo. 

Another issue I discovered is that Salvage can also use their breaching charges to blow open security crates(And also blow open the armory......). So now they don't spawn with bombs in their lockers, and need to pay for them with said mining points. 

<p>

image

</p>
</details>

# Changelog

:cl:
- add: Added a legality notice to the description of salvage's firearms. 
- tweak: The mining points cost of all salvage firearms have been increased by 4x
- remove: Salvage no longer starts with bombs in their lockers. Buy them with your mining points.

### DIFF
--- a/Resources/Locale/en-US/weapons/extenddescriptions/descriptions.ftl
+++ b/Resources/Locale/en-US/weapons/extenddescriptions/descriptions.ftl
@@ -1,0 +1,1 @@
+gun-legality-salvage = This weapon is licensed for use in planetary expeditions. It is unlawful to brandish this weapon aboard NanoTrasen stations, and may be confiscated by Security if removed from the station's Salvage Department.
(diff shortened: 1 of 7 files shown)


Erisfiregamer1 (approved):
A fair and balanced Salvage nerf. I like this, it actually makes sense as a Salvage player.

sleepyyapril (approved):

PR #1523, opened 2025-01-13
Maybe Fix Antag Related Crash

VMSolidus:
# Description

I'm trying to fix a crash caused by character profiles saving the same antag preference twice. I have no clue why or how it's happening, but this should probably prevent it from happening by making the profile sanitizer remove duplicate keys before saving them to the database.

The relevant crash log: 
[goobcrash.txt](https://github.com/user-attachments/files/18391763/goobcrash.txt)

# Changelog

:cl:
- fix: Attempted to fix a crash related to antag preferences saving multiple times.


VMSolidus:
Need @DEATHB4DEFEAT 's input

DEATHB4DEFEAT:
I think these are all being given as HashSets earlier in the stack but sure

PR #1528, opened 2025-01-13
AddTraitSpecial

Peptide90:
Adds a way to specify traits in the jobs special area.

Hashset so you can specify multiple traits. Tested on Nuclear14 to give the tribals a specific language in a friendly way.
```
  special:
  - !type:AddTraitSpecial
    traits: [ LanguageTribal, LanguageChinese ]
```

### DIFF
--- a/Content.Server/Jobs/AddTraitSpecial.cs
+++ b/Content.Server/Jobs/AddTraitSpecial.cs
@@ -4,35 +4,27 @@ using Content.Shared.Traits;
 using JetBrains.Annotations;
 using Robust.Shared.Prototypes;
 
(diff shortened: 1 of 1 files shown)


VMSolidus (approved):
I so desperately wished EntProtoId<t> actually worked in these, we otherwise could have skipped the need to index them in the first place. Otherwise yep, this is good here. I think you may be crazy here for using a hash set(Uniqueness, oddly enough, isn't actually required for traits). But fussing over which array type is best is an insane thing to ask for a function that isn't going to be iterating through tens of thousands at a time.

PR #1545, opened 2025-01-14
"Proper" "softcrit" "support"

RedFoxIV:
# Description

Implements the softcrit functionality.
Similiar to critical state but spessmen will be able to communicate and crawl around, but not pick up items.
Also supports configuring what is and isn't allowed in different MobStates (per mob prototype): you can enable picking up items while in softcrit so people can pick up their lasgun and continue shooting after taking a 40x46mm to their ass cheeks from the guest nukies while being dragged to safety.

escape-from-tarkov-raid

# Technical details
New prototype type: "mobStateParams" (`MobStateParametersPrototype`)
Used to specify what can and can't be done when in a certain mobstate. Of note that they are not actually bound to any `MobState` by themselves. To assign a params prototype to a mobstate, use `InitMobStateParams` in `MobStateComponent`.
It has to be a prototype because if I just did something akin to `Dictionary<MobState, Dictionary<string, bool>>`, you'd have to check the parent and copy every flag besides the one you wish to modify. That is, if I understand how the prototype system works correctly, which I frankly doubt. 

A new datafield serializer: `public class PrototypeIdDictionarySerializer<TPrototype> : ITypeValidator<Dictionary<string, string>, MappingDataNode> where TPrototype : class, IPrototype`
- meant to serialize `Dictionary<string, string>` with the key being some string and value being the id of `TPrototype` prototype.
 
MobStateComponent now has:
- `Dictionary<string, string> InitMobStateParams`, for storing "mobstate - parameter prototype" pairs. `<string, string>` because it has to be editable via mob prototypes. Named "mobStateParams" for mob prototypes.
- `public Dictionary<MobState, MobStateParametersPrototype> MobStateParams` for actually storing the params for each state
- `public Dictionary<MobState, MobStateParametersOverride> MobStateParamsOverrides` for storing overrides. `MobStateParametersOverride` is a struct which mirrors all `MobStateParametersPrototype`'s fields, except they're all nullable. This is meant for code which wants to temporarily override some setting, like a spell which allows dead people to talk. This is not the best solution, but it should do at first. A better option would be tracking each change separately, instead of hoping different systems overriding the same flag will play nicely with eachother.
- a shitton of getter methods

TraitModifyMobState now has:
- `public Dictionary<string, string> Params` to specify a new prototype to use.
- Important note: All values of `MobStateParametersPrototype` are nullable, which is a hack to support `TraitModifyMobState`. This trait takes one `MobStateParametersPrototype` per mobstate and applies all of its non-null values. This way, a params prototype can be created which will only have `pointing: true` and the trait can apply it (e.g. to critstate, so we can spam pointing while dying like it's a game of turbo dota)
- The above is why that wall of getters exists: They check the relevant override struct, then the relevant prototype. If both are null, they default to false (0f for floats.) The only exception is OxyDamageOverlay, because it's used both for oxy damage overlay (if null) and as a vision-limiting black void in crit..

MobStateSystem now has:
- a bunch of new "IsSomething"/"CanDoSomething" methods to check the various flags, alongside rewritten old ones.
- image lookin ahh predicate factory

---

# TODO

I am left dead inside after working on this because 80% of work was either writing boilerplate, replacing boilerplate with other boilerplate or fixing boilerplate that i broke with the cherrypick while moving softcrits to EE. Hence, this is a draft PR.

- [ ] Make an option to prohibit using radio headsets while in softcrit
- [ ] Make proper use of `MobStateSystem.IsIncapacitated()`.
Previous `IsIncapacitated()` implementation simply checked if person was in crit or dead. Now there is a `IsIncapacitated` flag in the parameters, but it's heavily underutilized. I may need some help on this one, since I don't know where would be a good place to check for it and I absolutely will not just scour the entire build in search for them.
- [ ] Don't drop items if `ForceDown` is true but `PickingUp` is also true.
- [ ] If above: make that another option? "DropItemsIfDowned"?
- [ ] probably move some other stuff to Params if it makes sense. Same thing as with `IsIncapacitated` though: I kinda don't want to, at least for now.
- [ ] Actually check what are "conscious attempts" are used for
- [ ] Look into adding a way to make people choke "slowly" in softcrit as opposed to choking at "regular speed" in crit. Make that into a param option? Make that into a float so the speed can be finetuned?
- [ ] make sure the serializer actually does its job
- [ ] Restore my lost sanity 
- [x] Make sure it at least compiles

---

<details><summary><h1>No media</h1></summary>
<p>

:p

</p>
</details>

---

# Changelog

:cl:
- add: Soft critical state. Crawl to safety, or to your doom - whatever is closer.


bruhmogus:
huh?? didnt we already have this???

RedFoxIV:
now that i am reading this a day later, I can confidently say this PR is borderline incomprehensible
wow

BlueHNT:
What the fuck happened here?

RedFoxIV:
> huh?? didnt we already have this???

`MobState.SoftCritical` was already present but went unused, as well as some other stuff relating to it. The closest to softcrit EE currently has is allowing people to crawl and/or talk while in crit.

> What the fuck happened here?

lack of QA

Erisfiregamer1:
> > What the fuck happened here?
> 
> lack of QA

Nothing in SS14 has Quality Assurance, my friend

RedFoxIV:
It's as ready as I will ever get it to be without someone else looking at it.

Content.Server/Traits/TraitSystem.Functions.cs
```diff
@@ -444,85 +448,29 @@ public sealed partial class TraitModifyMobState : TraitFunction
...
-            mobStateComponent.AllowHandInteractWhileDead = AllowHandInteractWhileDead.Value;
+        foreach (var pair in Params) {
+            DebugTools.Assert(_reflection.TryParseEnumReference($"enum.MobState.{pair.Key}", out var e), $"MobState.{pair.Key} does not exist.");
+            MobState state = (MobState) e;
+            MobStateParametersPrototype current = comp.MobStateParams[state];
+            var mod = _proto.Index<MobStateParametersPrototype>(pair.Value);
+            current.MergeWith(mod);
+            //current.FillDefaults();
+        }
+        comp.Dirty(); // why is this deprecated? it's much better than manually resolving entitymanager with ioc
```
VMSolidus, line 473:
```suggestion
        comp.Dirty(); // why is this deprecated? it's much better than manually resolving entitymanager with ioc
```
It's deprecated because it's been replaced with Dirty(uid, comp); instead of Dirty(comp). I think you might be able to get around the deprecated warning by doing comp.Dirty(uid);. The reason for the deprecation is because you're supposed to only dirty one component at a time, and if you dirty just the component, the system now has to dirty all of that component, so it needs UID + Component.

However we now have as an engine option, the ability to dirty **individual datafields** in components, feel free to take a look at that if you wish!

Content.Server/Traits/TraitSystem.Functions.cs
```diff
@@ -444,85 +448,29 @@ public sealed partial class TraitModifyMobState : TraitFunction
...
-        if (AllowHandInteractWhileCrit is not null)
-            mobStateComponent.AllowHandInteractWhileCrit = AllowHandInteractWhileCrit.Value;
-
-        if (AllowHandInteractWhileSoftCrit is not null)
-            mobStateComponent.AllowHandInteractWhileSoftCrit = AllowHandInteractWhileSoftCrit.Value;
-
-        if (AllowHandInteractWhileDead is not null)
-            mobStateComponent.AllowHandInteractWhileDead = AllowHandInteractWhileDead.Value;
+        foreach (var pair in Params) {
+            DebugTools.Assert(_reflection.TryParseEnumReference($"enum.MobState.{pair.Key}", out var e), $"MobState.{pair.Key} does not exist.");
```
VMSolidus, line 466:
```suggestion
            DebugTools.Assert(_reflection.TryParseEnumReference($"enum.MobState.{pair.Key}", out var e), $"MobState.{pair.Key} does not exist.");
```
You're not allowed to use e as a variable here because it straight up won't exist in Release configuration. Separate out TryParseEnumReference into an If Not {Assert; Continue;} statement to clear the error

Content.Shared/Mobs/Components/MobStateComponent.cs
```diff
@@ -14,92 +27,176 @@ namespace Content.Shared.Mobs.Components;
...
+    public bool IsThreatening(MobState? State = null) => GetOverride(State).Threatening ?? GetParams(State)?.Threatening ?? true; // :)
+    
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public bool CanEquipSelf(MobState? State = null) => GetOverride(State).CanEquipSelf ?? GetParams(State)?.CanEquipSelf ?? false;
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public bool CanUnequipSelf(MobState? State = null) => GetOverride(State).CanUnequipSelf ?? GetParams(State)?.CanUnequipSelf ?? false;
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public bool CanEquipOther(MobState? State = null) => GetOverride(State).CanEquipOther ?? GetParams(State)?.CanEquipOther ?? false;
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public bool CanUnequipOther(MobState? State = null) => GetOverride(State).CanUnequipOther ?? GetParams(State)?.CanUnequipOther ?? false;
```
VMSolidus, line 121: Holy fuck this is amazing. I'm taking notes.

VMSolidus (changes requested):
Holy shit this PR is amazing.

Content.Server/Traits/TraitSystem.Functions.cs
RedFoxIV: i am not having any errors here, but sure

RedFoxIV: the null checker actually stops seeing that e won't be null on the line 468, where it's cast into MobState var

Content.Server/Traits/TraitSystem.Functions.cs
RedFoxIV: not sure how its implementation changed over time, but currently `comp.Dirty()` is pretty much equivalent to `Dirty(comp.Owner, comp)`, barring a single resolve of EntityManager. I guess it's marked as obsolete because `Owner` field is obsolete too.
Being able to dirty individual datafields is, i suppose, about custom defined component states? First time i'm hearing about this.

Content.Shared/Mobs/Components/MobStateComponent.cs
RedFoxIV: :flushed:

RedFoxIV:
image
ain't my fault

BlueHNT:
Seems the issues is pool manager stuff.

stellar-novas:
We should really fix those tests.

VMSolidus (approved):

PR #1587, opened 2025-01-18
Split All CVars Into Separate Files

VMSolidus:
# Description

This PR splits all the CVars into separate files. Unfortunately I need to hella clean this up by restoring all the CVars that are different on EE, because cherrypicking it reset everything to Wizden's Defaults (Including CVars that no longer exist on EE, and removing CVars from systems I codeown now, such as Atmos, Guns, and Melee). 

# Changelog

No changelog because this isn't player facing. :)


VMSolidus:
I just now completed going through **every single CVar individually**, which took an entire hour, to ensure that all CVars are present, and exactly as-is on EE Master.

sleepyyapril (approved):

PR #1602, opened 2025-01-19
Consolidate Arsenal Research

VMSolidus:
# Description

Arsenal Research SUCKS. And the reason it sucks is twofold: Nothing it unlocks can be made by science directly. And it has entirely too many research unlocks that only grant 1 or 2 items max, which bloated the menu and meant that even if you wanted to pursue making some cool gun, you were punished for wanting to do so.

Well this PR does nothing to fix the first issue. But it does fix the second issue by consolidating a ton of the Arsenal research into larger single entries. Some things that were too basic were also removed as researchable outright, and made default in the security techfab (Basic technology that you could consider "Tier 0" that already exists on the station should NEVER be a research). 

# Changelog

:cl:
- tweak: Made Arsenal Research significantly more interesting by consolidating most of the useless single item researches into larger essential categories. Basic Weapons, Advanced Weapons, Prototype Weapons. Going forwards, these three categories will have significantly more exciting options added to them that should make it more worthwhile to pursue.


sleepyyapril (approved):

VMSolidus:
image

Stupid Heisentest...

PR #1607, opened 2025-01-20
Fix Possible Loadout Related Crash

VMSolidus:
# Description

Servers are occasionally experiencing roundstart crashes, and the logs I received concerning this crash imply that LoadoutSystem is SOMEHOW being handed a player character that doesn't actually exist. Well an easy solution I guess is to just make LoadoutSystem check if the entity *factually* exists, before attempting to apply a loadout to it.

# Changelog

:cl:
- fix: Possibly fixed a crash related to Loadouts being applied to entities that don't exist.


angelofallars (approved):

VMSolidus:
I want to see if it also yeets the Heisentest too.

PR #1624, opened 2025-01-21
Only Use Roundstart Species for RandomHumanoid

sleepyyapril:
it will no longer choose vox or lamia.

### DIFF
--- a/Content.Server/Humanoid/Systems/RandomHumanoidSystem.cs
+++ b/Content.Server/Humanoid/Systems/RandomHumanoidSystem.cs
@@ -1,3 +1,4 @@
+using System.Linq;
 using Content.Server.Humanoid.Components;
 using Content.Server.RandomMetadata;
 using Content.Shared.Humanoid.Prototypes;
@@ -19,6 +20,8 @@ public sealed class RandomHumanoidSystem : EntitySystem
 
     [Dependency] private readonly HumanoidAppearanceSystem _humanoid = default!;
 
+    private HashSet<string> NotRoundStartSpecies;
+
     /// <inheritdoc/>
     public override void Initialize()
     {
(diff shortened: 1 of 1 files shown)


angelofallars (approved):

Content.Server/Humanoid/Systems/RandomHumanoidSystem.cs
```diff
@@ -31,14 +34,26 @@ private void OnMapInit(EntityUid uid, RandomHumanoidSpawnerComponent component,
...
     public EntityUid SpawnRandomHumanoid(string prototypeId, EntityCoordinates coordinates, string name)
     {
         if (!_prototypeManager.TryIndex<RandomHumanoidSettingsPrototype>(prototypeId, out var prototype))
             throw new ArgumentException("Could not get random humanoid settings");
 
-        var profile = HumanoidCharacterProfile.Random(prototype.SpeciesBlacklist);
+        var blacklist = prototype.SpeciesBlacklist;
+
+        if (!prototype.SpeciesBlacklist.Any())
+            blacklist = _notRoundStartSpecies;
```
angelofallars, line 54: The blacklist should be a union of the `prototype.SpeciesBlacklist` hashset and the `_notRoundStartSpecies` hashset just in case the prototype disables more than the non-roundstart species.

angelofallars (changes requested):

Content.Server/Humanoid/Systems/RandomHumanoidSystem.cs
sleepyyapril: You read it wrong, I ignore the notRoundStartSpecies when there is a SpeciesBlacklist with any values because they should be able to choose on their own.

angelofallars (approved):

PR #1639, opened 2025-01-23
Remove Gamerule Count Test 

sleepyyapril:


VMSolidus (approved):
Makes me wonder what Wizden does with these? Do they just not have sub gamerules anymore?

sleepyyapril:
im just early merging because it's just commenting out tests

PR #1645, opened 2025-01-23
Removes Asphyxiation From Lamia in Favor of Pierce/Poison

BlueHNT:
# Description

The asphyxiation damage was extremely temporary and it took less than a cycle to get it healed up. In the current stage where you can't strangle someone, this kind of damage is irrelevant.

Thus I have moved the 2 points of asphyx equaly to piercing and poison instead, making the lamia melee feel a bit stronger.

---

# Changelog

:cl:
- tweak: Tweaked lamia bite

### DIFF
--- a/Resources/Prototypes/DeltaV/Entities/Mobs/Species/lamia.yml
+++ b/Resources/Prototypes/DeltaV/Entities/Mobs/Species/lamia.yml
@@ -167,9 +167,8 @@
     animation: WeaponArcBite
     damage:
       types:
-        Piercing: 1
-        Poison: 2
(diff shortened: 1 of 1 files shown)


BlueHNT (commented):
There's still a few things I want to try before putting this up for ready

VMSolidus (approved):
Yea when I originally started making them, poison was basically "Not possible to heal without going to medbay", I was definitely mistaken in thinking asphyix was viable as a damage type. And poison's threat is a bit diminished now by there being two different traits that heal it passively

PR #1657, opened 2025-01-25
Better Teach Lesson Code

VMSolidus:
# Description

Thanks to Router for pointing out we can actually reuse code that is apparently just always running exclusively on player characters. So it turns out that this game actually has a system for tracking who the killer is. We can skip enumeration outright if the killer was the one who downed the player. There's still BETTER ways to do this, like subscribing to KillReportedEvent instead of MobStateChanged, but I'm a bit too zonked out to pursue that right now.


sleepyyapril (approved):

VMSolidus:
image
Reeeeeeee

PR #1661, opened 2025-01-26
Make Markings Less Hell + Chitinid Hotfix

sleepyyapril:
i probably missed a few but someone better than me with markings can fix 'em later


VMSolidus (approved):
YE

PR #1665, opened 2025-01-26
Mind Over Machine, Machine Over Mind

VMSolidus:
# Description

I'm going to go put powergamers on suicide watch. This PR makes it so that specializing in psionic traits, and specializing in cybernetic traits are both mutually exclusive. You are only allowed to have a single cybernetric trait if you wish to have psionic traits. And you are only allowed to have Latent Psychic with no other traits if you wish to have Cybernetics. 

Also fixes a bug with Thermographic Vision not correctly being measured in seconds. You now get a 2 second pulse with Thermographic vision, as intended.

# Changelog

:cl:
- tweak: Psionic traits are now mutually exclusive with cybernetic traits, and vice versa.
- fix: Thermographic Vision now correctly measures its pulse duration in seconds instead of nanoseconds. It provides a 2 second scan.


VMSolidus:
image

Utterly baffling that the suicide test can pick a species that is disabled.

VMSolidus:
image

Tested that these do indeed work, making this PR the first actual ingame use of ItemGroups for Traits.

sleepyyapril:
It already takes a lot of penalty to take _both_, so I really don't recommend this.

VMSolidus:
I feel like both of my maintainers here didn't at all understand what I'm saying. Which is on me.

Problem: I can get 24 *extra* trait points, at no downside whatsoever, by using 7 of my trait selections to buy latent psychic + all 6 psionic negative trait. These negative traits **only matter if you are specialized in psionics**. If I have completely zero intention to ever interact with psionics, these negatives will never do what they are supposed to be doing. 

Problem: With 34 trait points and 7 selections, you can buy Dermal Armor + Bionic Arm + Platelet Factories, and still have points leftover for more. All 3 of those traits are priced at the assumption that a player should not have more than 1 unless they are heavily laden down by negative traits.

**Notice**: You are still allowed ONE trait from the opposite category. You can be a wizard with dermal armor. OR you can be a cyborg with an ever so faint hint of psychic potential.

**Notes**: Even Shadowrun and Cyberpunk Red BOTH used limitations just like this for exactly this reason. Shadowrun has, "The DM needs to inspect your character for approval before game start to make sure you aren't trying to pay for Wizard Positives with Hacker Negatives". Your Wizard will literally never jack into the matrix, so they shouldn't be allowed to take negative traits that only come up for hackers.

DEATHB4DEFEAT (approved):

stellar-novas:
Ahh, that makes significantly more sense than what I thought.

stellar-novas (approved):

PR #1677, opened 2025-01-28
New Solar Sprites, New Solar Panel Upgrades, and some Solar Panel Fixes

sleepyyapril:
port https://github.com/space-wizards/space-station-14/pull/29224

:cl: CaasGit
- add: Solar panels can now be upgraded with plasma or uranium glass.


VMSolidus (changes requested):
Need to literally quadruple the power stats of all 3, because these wizden numbers are still garbage even "upgraded". Keep in mind we had previously 4x'd them, but wizden didn't, so the uranium panels are half the power output of buffed panels on EE.

VMSolidus (approved):

PR #1726, opened 2025-02-06
Give BSO Access to HoS Energy Sword (for Now)

EctoplasmIsGood:
# Description

I have no idea what I'm doing I'm sure this will all work out. Literally all it does is gives BSO access to the HOS Energy Sword because I think BSO is underwhelming in terms of combative abilities, and needs a melee option.

Also the HoS energy sword is blue and I think that would be cool for BSO to have until they have their own unique weapons.

---

# Changelog

- Add: Added HoS energy sword to BSO loadout

### DIFF
--- a/Resources/Prototypes/Loadouts/Jobs/Command/blueshield_officer.yml
+++ b/Resources/Prototypes/Loadouts/Jobs/Command/blueshield_officer.yml
@@ -47,6 +47,20 @@
 # Ears
 
 # Equipment
+# Blueshield Officer Weapon Selection
+# Currently just stolen from HOS until BSO gets its own unique weapon(s) 
+# There was no group for BSO weapons so I just removed the requirement, should be fine for now.
+- type: loadout
+  id: LoadoutCommandHoSEnergySword
+  category: JobsCommandBlueshieldOfficer
+  cost: 0
+  canBeHeirloom: true
+  requirements:
+   - !type:CharacterJobRequirement
+      jobs:
+        - BlueshieldOfficer
+  items:
+    - EnergySwordHoS
 
 # Eyes
 


EctoplasmIsGood:
I probably fucked up something very simple somewhere because I've never done this and went in blind.

SixplyDev:
Not a fan if I'm being honest, BSO having *three* weapons especially with the fact he's not limited at all. There's melee options with your guns & I feel like those are fine.

EctoplasmIsGood:
> Not a fan if I'm being honest, BSO having _three_ weapons especially with the fact he's not limited at all. There's melee options with your guns & I feel like those are fine.

The chester is a mediocre clip fed pump/lever action gun. It does decent damage per shot but its fire rate makes it abyssmal in actual combat, and firespeed that slow cannot compete against a lot of common weapons. Energy revolver is okay, but acts as a sidearm and is typically used as a glorified disabler. I believe the HoS esword is a temporary bandaid fix to give BSO an edge in combat until it gets actual custom tailored weapons.

Chester bad, BSO weak, make BSO slightly stronger, very good.

VMSolidus (changes requested):
>Giving BSO the HoS' energy sword for free
Absolutely not. For one it's a *Unique traitor objective item* of which only one is supposed to exist. 

You'd be better off letting the BSO choose their sidearm from the same list as regular security officers.

EctoplasmIsGood:
> * Giving BSO the HoS' energy sword for free
> 
> Absolutely not. For one it's a _Unique traitor objective item_ of which only one is supposed to exist.
> 
> You'd be better off letting the BSO choose their sidearm from the same list as regular security officers.

I was completely unaware it was an antag requirement to be honest. And while giving them access to secoff loadouts is an option, it still feels like theyre less armed then they should be. I'll close this though with that new knowledge.

PR #1730, opened 2025-02-06
BSO Lifeline Port

Sarahon:
# Description

BSO lifeline implant - ported from Goobstation.

What is it?: It's a bluespace implant. Triggers when you die or when you click the button twice.

Why is it?: BSO is a special role, specifically employed by CC. They exist on station only to protect to command and nothing else. They have pretty strong armour and weapons. So this is the solution to that + creates an RP narrative.

It spawns in the BSO's pocket, as it's an optional implant, should probably be taken either start of shift or when things get dicey for command.

Additional: Adds new sprites for CC specified implants. 

---

<p>

Content Client_MHAiLxtLug

</p>
</details>

---

# Changelog
:cl:
- add: Blueshield Officer lifeline implanter


Resources/Prototypes/_Goobstation/Objects/Misc/subdermal_implants.yml
```diff
@@ -0,0 +1,21 @@
+- type: entity
+  parent: BaseSubdermalImplant
+  id: BluespaceLifelineImplant
+  name: bluespace lifeline implant
+  description: This implants teleport user body to safe place, forcing him to leave current location so he won't be able to come back.
```
angelofallars, line 5: Grammar:
`This implant teleports the user's body to a safe place, forcing them to leave the current location so they won't be able to come back.`

angelofallars (changes requested):

Resources/Prototypes/_Goobstation/Objects/Misc/subdermal_implants.yml
Sarahon: It was directly ported, I'll fix it in a second.

sleepyyapril:
what is this? link to the original PR or provide a description please

Sarahon:
> what is this? link to the original PR or provide a description please

Original PR is deleted. The media example shows you exactly what it does.

Resources/Prototypes/_Goobstation/Roles/Jobs/Command/blueshield_officer.yml
```diff
@@ -68,6 +68,7 @@
     gloves: ClothingHandsGlovesCombat
     id: BlueshieldPDA
     pocket1: UniqueBlueshieldOfficerLockerTeleporter
+    pocket2: BluespaceLifelineImplanter
```
VMSolidus, line 71: Maybe I'm crazy, but I could have swore we had a JobSpecial function that adds an implant to a job upon spawning in? Is it intended that the BSO doesn't spawn with it already implanted, and instead has it in their pocker?

angelofallars: You wouldn't want to put this implant in all situations, as it round-removes you when you die.

Content.Server/_Goobstation/Explosion/Components/GoobTriggerSystem.cs
```diff
@@ -0,0 +1,23 @@
...
+
+        SubscribeLocalEvent<DeleteParentOnTriggerComponent, TriggerEvent>(HandleDeleteParentTrigger);
+    }
+
+    private void HandleDeleteParentTrigger(Entity<DeleteParentOnTriggerComponent> entity, ref TriggerEvent args)
+    {
+        if (!TryComp<TransformComponent>(entity, out var xform))
+            return;
+
+        EntityManager.QueueDeleteEntity(xform.ParentUid);
```
VMSolidus, line 20: The TryComp isnt necessary since this is in EntitySystem, and Transforms are cached for ease of access. You can delete the TryComp entirely, and replace xform.ParentUid with Transform(entity).ParentUid

However that isn't why I'm requesting changes, because **OH MY GOD** this function is basically a disaster waiting to happen. Please for the love of god put in a safety check to prevent it from deleting Grids or Maps. Check if the parent is one of those two and return if either is the case. All it would take for a disaster to occur is for someone to introduce a way to "Activate" objects, and for it to activate an unused implant parented to the Station, and poof goes the station.

VMSolidus (changes requested):

Content.Server/_Goobstation/Explosion/Components/GoobTriggerSystem.cs
Aidenkrz: im 90% sure if this did somehow happen it would delete the implanter, not the station, because the implant is parented to the implanter but yeah

Resources/Prototypes/_Goobstation/Roles/Jobs/Command/blueshield_officer.yml
Sarahon: Yeah I'd prefer it stay optional vs forced because it's more a situational item. Not a must.

Content.Server/_Goobstation/Explosion/Components/GoobTriggerSystem.cs
Sarahon: > The TryComp isnt necessary since this is in EntitySystem, and Transforms are cached for ease of access. You can delete the TryComp entirely, and replace xform.ParentUid with Transform(entity).ParentUid
> 
> However that isn't why I'm requesting changes, because **OH MY GOD** this function is basically a disaster waiting to happen. Please for the love of god put in a safety check to prevent it from deleting Grids or Maps. Check if the parent is one of those two and return if either is the case. All it would take for a disaster to occur is for someone to introduce a way to "Activate" objects, and for it to activate an unused implant parented to the Station, and poof goes the station.

Ok so i've changed it to allow for any component with INonDeletable to not be affected by the system. So I could just add it to the grid component but durk doesn't seem to agree with this. Not sure of a better way after this.

Sarahon: Alright, I think I got it there.

VMSolidus (approved):

PR #1734, opened 2025-02-06
Exorcism of the Spirit of the Dwarf Species

trashalice:
# Description

The dwarf species has not been enabled in however long, removed them from the guidebook as they are no longer relevant. Use custom species to make a dwarf.

---

# Changelog

:cl:
- Remove: Dwarves from guidebook, begone beer boy.


stellar-novas (approved):

VMSolidus (approved):
Goodbye dwarfs. Maybe you'll return if we ever get around to doing the Dwarf Rework. That won't happen until we can get a proper Rock And Stone button.

PR #1735, opened 2025-02-07
Fix Shop Vendor Buying on Non-Click (#2906)

sleepyyapril:
:cl:
- fix: Salvage vendors will no longer accept any inputs to buy an item.


VMSolidus (approved):
Huh, interesting that it didn't previously check for this.

PR #1765, opened 2025-02-11
Space Wind Version 5 (FINAL)

VMSolidus:
# Description

MY ASCENSION IS COMPLETE, I HAVE DEFEATED THE ATMOS SYSTEM. MY POWERS HAVE GROWN TO NEW AND UNFATHOMABLE HEIGHTS. THE EQUATIONS FOR SPACE WIND ARE NOW PERFECTED. I HAVE BECOME THE PHYSICS CHAD THAT CAN NOW ACCOUNT FOR FRICTION AND AIR RESISTANCE. 

SPACE WIND HAS BEEN PERFECTED. THERE IS NO MORE WORK FOR ME TO DO ON IT.

# TODO

<p>

PROOF THAT MY EQUATIONS HAVE BEEN PERFECTED.

https://github.com/user-attachments/assets/b3eb43b2-e8db-4f75-a142-2c86308e8113

</p>
</details>

# Changelog

:cl:
- add: Added Space Wind V5. This version of space wind now accounts for the coefficients of friction and gravity as a calculated way to resist space wind. There is no more guesswork on if space wind should be able to throw an object. Additionally, thanks to a standalone module from Matrix Airflow System, space wind is capable of calculating airflow at arbitrarily any point in a room, regardless of whatever garbage Monstermos is saying.


Content.Shared/CCVar/CCVars.Atmos.cs
```diff
@@ -10,8 +10,17 @@ public sealed partial class CCVars
...
         CVarDef.Create("atmos.space_wind", true, CVar.SERVERONLY);
 
+    /// <summary>
+    ///     Space Wind with "Full Float Precision", allowing for curved motion of air.
+    ///     This is provided as a "Standalone" version of SWV4 that does not require the full MAS System.
+    ///     But as a consequence of being "Bolted on top of" Monstermos, it has no time complexity savings.
+    ///     THIS IS SIGNIFICANTLY MORE EXPENSIVE THAN EVEN WIZDEN'S SPACE WIND. YOU ARE WARNED.
+    /// </summary>
+    public static readonly CVarDef<bool> SpaceWindV4 =
+        CVarDef.Create("atmos.space_wind_version_4", true, CVar.SERVER);
```
Aidenkrz, line 20: "version 4" is a bad name for this. just call it expensive space wind

Aidenkrz (changes requested):

Content.Shared/CCVar/CCVars.Atmos.cs
sleepyyapril, line 20:
```suggestion
        CVarDef.Create("atmos.space_wind_version_4", false, CVar.SERVER);
```

sleepyyapril (changes requested):
default should be less expensive

VMSolidus:
After doing a bit more digging and talking with people, I figured out there's still a few more optimization stuffs I need to do. For one SpaceWindV4 does not require, nor benefit at all from MonstermosUseExpensiveAirflow. And I also learned that being made to use Config.GetCVar was a direct cause of significant latency issues in atmos. So when I get home I'll be swapping every Atmos CVar I've ever made over to Subs.CVar => x. The former actually has *substantial* performance cost in physics systems like these, and was a significant reason why actual server data wasn't lining up with what time and CPU instruction complexity models were indicating. 

I would like to do some testing with V4 fully replacing V3, as well as UseExpensiveAirflow removed outright. If its performance impact after making the appropriate fixes is less than that of the current data from Goob MRP, then I would strongly prefer that V4 be the default. 

I also need to move the MAS files into Shared, since going forward everything to do with MAS needs to be Shared to begin with so that we can build up to the holy grail of "Client predicted atmos".

VMSolidus:
I've worked out a proof of a solution to the "Wall Problem"(throw directions being incorrect if one or more walls are nearby), which requires only 8 more addition operations, so an incredibly insignificant increase to time complexity. It'll also fix the "Center tile doesnt throw" bug.

I'm putting it in tonight when I get home.

Content.Shared/CCVar/CCVars.Atmos.cs
BlueHNT: Is the expensive part gonna be toggleable then? On/Off cvar would be good either way for it.

VMSolidus:
Let's hear from my friend, the Biesel Republic navy officer!

@coderabbitai full review

VMSolidus:
@coderabbitai full review

VMSolidus:
I HAVE DEFEATED THE SPACE WIND SYSTEM. MY VICTORY OVER IT IS NOW FINISHED.

VMSolidus:
I was also able to find a very large set of optimizations for its interactions with the Throwing System, by caching a lot of information that the throwing system would have asked for to begin with, immediately before enumerating "Every active tile", such that the throwing system doesn't need to enumerate "Every projectile" once per item per tile.

Erisfiregamer1:
Holy shit he did it

BlueHNT:
I don't get it. Wouldn't the pressure in the tiles inside immediately funnel out to equalize the pressure? All I see is initial close objects getting sucked out and the rest spinning around pretty much freely.

Content.Server/Atmos/EntitySystems/AtmosphereSystem.HighPressureDelta.cs
```diff
@@ -154,29 +152,39 @@ private void ConsiderPressureDifference(GridAtmosphereComponent gridAtmosphere,
...
         TransformComponent? xform = null,
         PhysicsComponent? physics = null)
     {
         var (uid, component) = ent;
         if (!Resolve(uid, ref physics, false)
             || !Resolve(uid, ref xform)
             || physics.BodyType == BodyType.Static
-            || float.IsPositiveInfinity(component.MoveResist))
+            || float.IsPositiveInfinity(component.MoveResist)
+            || physics.LinearVelocity.Length() >= SpaceWindMaxVelocity)
```
BasedUser, line 170: Doesn't this prevent higher velocity objects (even if they're moving *against* the suggested wind direction) from being affected by pressure difference?

VMSolidus:
I had the Physics Professor at my college look at this, and he was amazed by how realistic the model. Between that and Router's seal of approval, this is good to go. :)

Content.Server/Atmos/EntitySystems/AtmosphereSystem.MAS.cs
```diff
@@ -0,0 +1,50 @@
...
+                || tileAtmosphere.Air is null
+                || tileAtmosphere.AirtightData.BlockedDirections is AtmosDirection.All
+                || tileAtmosphere.AirtightData.BlockedDirections.HasFlag(dir))
+                continue;
+
+            var pressureDiff = tile.Air!.Pressure - tileAtmosphere.Air.Pressure;
+            pressureVector += new Vector2(x, y) * pressureDiff;
+        }
+        return pressureVector;
+    }
```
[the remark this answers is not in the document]
Remuchi: AirtightData cannot be null

Content.Shared/Throwing/ThrowingSystem.cs
```diff
@@ -51,6 +51,7 @@ public override void Initialize()
         Subs.CVar(_configManager, CCVars.TileFrictionModifier, value => _frictionModifier = value, true);
     }
 
+    [Obsolete("If you are foreaching every entity, go get a ProjectileQuery and use TryThrow with EntityQuery<ProjectileComponent> instead. Otherwise, put it under pragma ignore.")]
```
Remuchi, line 54: I'd prefer making it a remark rather than making it an obsolete method

Content.Shared/Throwing/ThrowingSystem.cs
```diff
@@ -83,6 +84,7 @@ public void TryThrow(
     /// <param name="friction">friction value used for the distance calculation. If set to null this defaults to the standard tile values</param>
     /// <param name="compensateFriction">True will adjust the throw so the item stops at the target coordinates. False means it will land at the target and keep sliding.</param>
     /// <param name="doSpin">Whether spin will be applied to the thrown entity.</param>
+    [Obsolete("If you are foreaching every entity, go get a ProjectileQuery and use TryThrow with EntityQuery<ProjectileComponent> instead. Otherwise, put it under pragma ignore.")]
```
Remuchi, line 87: same as here

Remuchi (approved):
looks fine to me

Content.Server/Atmos/EntitySystems/AtmosphereSystem.MAS.cs
```diff
@@ -0,0 +1,50 @@
...
+        var pressureVector = new Vector2(0, 0);
+        foreach (var (x, y, dir) in MASSearchPattern)
+        {
+            if (!gridAtmos.Tiles.TryGetValue(tile.GridIndices + (x, y), out var tileAtmosphere)
+                || tileAtmosphere.Air is null
+                || tileAtmosphere.AirtightData.BlockedDirections is AtmosDirection.All
+                || tileAtmosphere.AirtightData.BlockedDirections.HasFlag(dir))
+                continue;
+
+            var pressureDiff = tile.Air!.Pressure - tileAtmosphere.Air.Pressure;
```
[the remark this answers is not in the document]
VMSolidus: Our compiler is kind of dumb about these, and doesn't understand the nullable variable can't be null if guarded by a continue rather than a return. So it'll refuse to compile if I don't have that ! operator there.

PR #1772, opened 2025-02-12
Derotate Unmaintained Maps.

VMSolidus:
# Description

We have a very large number of maps that outright have not received any meaningful updates in over a year. All of them being maps that we inherited from DeltaV during the initial forking. I've spent the entire past year trying to find people willing to maintain them, but to no avail. The rot is incredibly noticeable, and is particularly egregious when they get compared to maps being maintained by SiN Mapping Team and Estacao Pirata. We can't guarantee the quality of these maps, and the original maintainers from DeltaV have expressed zero desire to offer any assistance with maintaining them. 

I put it to a vote last night on the EE Discord, in order to find out which of the DeltaV maps were actually worth keeping, and the vote was overwhelmingly in favor of Shoko, Glacier, and Lighthouse. I'm willing to work with my existing map maintainers to focus on getting these three maps the care and attention they need in order to be on par with the quality expectation that maps like Gax and Lambda managed to set. 

If we want more maps, we need people willing to maintain the maps. These maps aren't being removed from the game, just set aside and essentially marked as "Unsupported". If in the future someone wishes to step forward and claim responsibility for one of the maps, they can submit a PR to re-rotate the map in question, assuming it is paired with *some* amount of work done to refurbish the map. It doesn't need to be a full rework, just enough work to prove the map is being taken care of.

# Changelog

:cl:
- tweak: Removed the following maps from rotation: Pebble, Edge, Tortuga, Arena, Asterisk, Hive, Hammurabi, and Submarine. These maps aren't permanently gone, they just won't appear in the map vote list. If you want to see them return in an official capacity, and have any talent/desire for mapping, feel free to submit a PR to give one of these maps the attention and care it deserves. None of them have received any updates in over a year.

### DIFF
--- a/Content.IntegrationTests/Tests/PostMapInitTest.cs
+++ b/Content.IntegrationTests/Tests/PostMapInitTest.cs
@@ -52,27 +52,27 @@ namespace Content.IntegrationTests.Tests
             "CentCommHarmony",
             "MeteorArena",
             "NukieOutpost",
-            "Core",
-            "Pebble", //DeltaV
-            "Edge", //DeltaV
-            "Saltern",
-            "Shoukou", //DeltaV
-            "Tortuga", //DeltaV
-            "Arena", //DeltaV
-            "Asterisk", //DeltaV
-            "Glacier", //DeltaV
-            "TheHive", //DeltaV
-            "Hammurabi", //DeltaV
-            "Lighthouse", //DeltaV
-            "Submarine", //DeltaV
-            "Gax",
-            "Rad",
-            "Europa",
-            "Meta",
-            "Box",
-            "Lambda",
-            "Bagel",
-            "Northway"
+            "Core", // No current maintainer. In need of a rework...
+            // "Pebble", // De-rotated, no current maintainer.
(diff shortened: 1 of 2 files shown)


angelofallars (approved):

OldDanceJacket:
Worth noting is that maps like Hive are made for high pop anyways so votes will be skewed due to that.

OldDanceJacket (approved):

VMSolidus:
> Worth noting is that maps like Hive are made for high pop anyways so votes will be skewed due to that.

Hive for example is definitely worth refurbishing, it's not even a terrible map. It could almost certainly could just do with a "Detailing Pass" at most, maybe make the AI Satellite on it suck less, which isn't hard to do.

OldDanceJacket:
It's really an optics thing as well, Lighthouse is a fundamentally flawed/poor map in the same way that Bagel is, but is a 'learned' station that you can comfortably play on, Glacier is simply an aesthetics thing as the map is also really bad, if a better designed "station on top of a cold mountain" were to walk around Glacier would be given the boot in a heartbeat.

Shoko is a good map that works for low or mid mop however.

PR #1774, opened 2025-02-12
Planet Maps Prep For Glacier Rework

VMSolidus:
# Description

Goddamn do I love Adderall.

This PR lays the entire groundwork needed for the "Glacier Rework", and it does so by fixing every problem I could think of between FTL and StationBiome that was preventing it from being a reality. Essentially, I'm hitting several birds with one stone by refactoring both systems to be a LOT more flexible. You can technically say that this basically is also partly enabling Lavaland maps, since I conveniently included options in the StationBiomeComponent to generate dungeons. Technically a planet map should probably also have a RestrictedRangeComponent added to it so that players can't wander so far from the station that they generate a truly excessive number of entities. 

Yes, this does infact mean you can now run Nukie gamemodes on planet maps. 

<p>

https://github.com/user-attachments/assets/7a3730af-c521-42d4-8abd-5aa5989c369c

</p>
</details>

---

# Changelog

:cl:
- add: Shuttles can now take off and land from planet surfaces. They however will still respect exclusion zone beacons set by stations. Maybe in the future there might be a special shuttle (drop pod) that is set to ignore these exclusion zones.

### DIFF
--- a/Content.Server/Station/Systems/StationBiomeSystem.cs
+++ b/Content.Server/Station/Systems/StationBiomeSystem.cs
@@ -1,9 +1,12 @@
 using Content.Server.Parallax;
+using Content.Server.Procedural;
 using Content.Server.Station.Components;
 using Content.Server.Station.Events;
 using Content.Shared.Parallax.Biomes;
 using Robust.Server.GameObjects;
+using Robust.Shared.Map.Components;
 using Robust.Shared.Prototypes;
+using Robust.Shared.Random;
 
 namespace Content.Server.Station.Systems;
 public sealed partial class StationBiomeSystem : EntitySystem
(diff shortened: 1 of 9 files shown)


Content.Server/Station/Systems/StationBiomeSystem.cs
```diff
@@ -28,8 +33,32 @@ private void OnStationPostInit(Entity<StationBiomeComponent> map, ref StationPos
         if (station == null) return;
 
         var mapId = Transform(station.Value).MapID;
-        var mapUid = _mapManager.GetMapEntityId(mapId);
+        if (!_mapSystem.TryGetMap(mapId, out var mapUid)
+            || mapUid is null) // WHAT, IT'S LITERALLY CALLED TRYGET. WHY DO I NEED TO CHECK NULL?
```
DEATHB4DEFEAT, line 37: Tried to see why this happens but opening rider killed my computer

DEATHB4DEFEAT (approved):

PR #1775, opened 2025-02-13
Disposable Syringes

Timfa2112:
# Description

A cheap disposable syringe. It can draw once, inject once, and is then rendered unusable.
The point of the syringe is to let Chemists pack specific dosages of specific chemicals into a simple syringe. A doctor can then not accidentally alter a dosage during treatments.

These cheap syringes live alongside normal syringes. They use fewer materials to make, and are printed much faster.

---

<p>

https://github.com/user-attachments/assets/eb439050-b86d-49ba-b95e-22ab271f2358

https://github.com/user-attachments/assets/9e8954ec-11c9-4569-820e-08b91e09f52b

</p>
</details>

---

# Changelog

:cl:
- add: Added disposable syringes! These cheaper, quicker-to-produce syringes let Chemists pick dosages while preparing medicine so that doctors can inject without worrying about volume, but the syringes cannot be used more than once.


stellar-novas (commented):
Could they be a chemmaster product too? Similar to pills?

Timfa2112:
> Could they be a chemmaster product too? Similar to pills?

I originally considered this, but then noticed it would really make them no different from pills, which seemed unnecessary. I wouldn't want to replace pills, after all.

Content.Client/Chemistry/UI/FillableOneTimeInjectorStatusControl.cs
```diff
@@ -0,0 +1,84 @@
...
+            else
+            {
+                _label.SetMarkup(
+                    Loc.GetString(
+                        "injector-volume-label",
+                        ("currentVolume", solution.Volume),
+                        ("totalVolume", solution.MaxVolume),
+                        ("modeString", modeStringLocalized),
+                        ("transferVolume", _parent.Comp.TransferAmount)));
+            }
```
sleepyyapril, line 82: fix indentation, one tab too far

Content.Shared/Chemistry/EntitySystems/SharedFillableOneTimeInjectorSystem.cs
```diff
@@ -0,0 +1,108 @@
...
+using Robust.Shared.Player;
+
+namespace Content.Shared.Chemistry.EntitySystems;
+
+public abstract class SharedFillableOneTimeInjectorSystem : EntitySystem
+{
+    /// <summary>
+    ///     Default transfer amounts for the set-transfer verb.
+    /// </summary>
+    public static readonly FixedPoint2[] TransferAmounts = { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15 };
```
sleepyyapril, line 19:
```suggestion
```

Content.Shared/Chemistry/EntitySystems/SharedFillableOneTimeInjectorSystem.cs
```diff
@@ -0,0 +1,108 @@
...
+            },
+
+            Priority = priority
+        };
+        args.Verbs.Add(toggleVerb);
+
+        priority -= 1;
+
+        // Add specific transfer verbs according to the container's size
+        foreach (var amount in TransferAmounts)
```
sleepyyapril, line 67:
```suggestion
        for (var amount = 0; amount < 15; amount++)
```

sleepyyapril (changes requested):

Content.Shared/Chemistry/EntitySystems/SharedFillableOneTimeInjectorSystem.cs
sleepyyapril: don't blindly copy this, it's a web review lmao

sleepyyapril (approved):

sleepyyapril:
you can just do amount++ instead of amount += 1 but it's whatever

Timfa2112:
> you can just do amount++ instead of amount += 1 but it's whatever

Not with a FixedPoint2 I think, it complained when I tried that

PR #1777, opened 2025-02-14
Re-Add Missing Recipes To Industrial Ore Processor

angelofallars:
# Description

Re-adds the missing recipes to the industrial ore processor which were removed by accident in https://github.com/Simple-Station/Einstein-Engines/pull/1737.

Also adds a missing localization string for `materials-raw-diamond`.

## Media

**Updated industrial ore processor recipes**

**Missing localization**

## Changelog

:cl: Skubman
- fix: The industrial ore processor can now make plasteel, uranium glass, reinforced uranium glass, and refined diamonds again.


sleepyyapril (approved):

Resources/Prototypes/Entities/Structures/Machines/lathe.yml
```diff
@@ -1515,6 +1515,29 @@
   - type: Lathe
     materialUseMultiplier: 0.75
     timeMultiplier: 0.5
+    staticRecipes:
+    - BluespaceCrystal
```
VMSolidus, line 1519: Remove this entire list from the T2 Ore Processor, and add every missing material to the Base Ore Processor.

VMSolidus (changes requested):

VMSolidus:
To clarify, I'm opposed to "Different Tiers" of lathe machines having different recipe lists, because if at some point we introduce more tiers of lathe machines and various upgraded forms of them(This isn't even that outlandish. Consider that Science research isn't actually limited to T3), then that means maintaining even more lists of recipes. The right way to go about this is to only add recipes to the Base of a specific lathe type. So that for instance, "Security Techfab" contains the majority of all Arsenal Research items, and if we add different tiers of Security Techfabs(This is actually something I want to do. I want a Hyperconvection Secfab for instance), the tier 2 version of it only needs to parent the recipe list from the standard one. 

That's why I'm requesting that the missing recipes from T2 Ore Processor be instead put in T1 Ore Processor, and for the T2 Ore Processor to just inherit its recipes from its parent.

PR #1785, opened 2025-02-14
Telecrystal Rebalance

VMSolidus:
# Description

This brings over the "Telecrystal Rebalance". To make the TC costs more granular, nearly all values for TC costs have been multiplied by 5x, including the starting TC amounts for each type of uplink. Additionally, Nukies no longer spawn with any of their tacsuits, and have had the cost of said tacsuits refunded as 50 additional TC. Finally, since the Commander and Medic no longer spawn with their respective tacsuits, they've been put into the uplink as new bundles(which traitors can buy as well).

I figure its about time the nukies get some love seeing as on MRP they're facing against Security who can buy an N1984 in their loadouts. Next PR I'm probably going to even further mess with the cost of guns by separating out the guns from their magazines, and adding options for buying more "Special Ammo" types to replace them. Such as for instance, a "Rubber Ammo Bundle" for Nukies, or a more expensive "Incendiary Ammo Bundle". 

# Changelog

:cl:
- tweak: Added the "Telecrystal Rebalance" from Goob.

### DIFF
--- a/Resources/Prototypes/Catalog/Fills/Backpacks/duffelbag.yml
+++ b/Resources/Prototypes/Catalog/Fills/Backpacks/duffelbag.yml
@@ -284,6 +284,36 @@
 
 - type: entity
   parent: ClothingBackpackDuffelSyndicateBundle
+  id: ClothingBackpackDuffelSyndicateHardsuitCommanderBundle
+  name: syndicate commander hardsuit bundle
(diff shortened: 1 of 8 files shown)


sleepyyapril (approved):

sleepyyapril:
reminds me that we have a few uplink related bugs, e.g. traitors sometimes not given codewords. maybe a wizden PR exists?

PR #1795, opened 2025-02-14
BSO Paramed Gaming

EctoplasmIsGood:
# Description

BSO as of now is treated as a purely combatant role, and while it is for the most part, I see it being closer to Corpsman. This PR adds new medical options to the BSOs loadout so that they can keep their charges in good health. It can be debated that this is overturned/power creep, and you're right to an extent. BSO is currently in a weird spot, as a lot of people disagree on how it should function, however, I believe that giving them access to medical items will lock in their position as the guardian/caretaker of command, and will overall be good for their kit and performing their jobs duties. 

BSO is the Corpsman of command, and them having access to high-tier equipment to match that makes sense. The crew monitor is a very good item, yes, but due to the fact that BSO will ONLY be using these items on command, I believe that them having it won't disrupt balance too much. 

(Crew monitor makes it possible for BSO to actually keep track of the entirety of command, rather then sitting on one all game and praying they respond to comms)

<p>

Example Media Embed

</p>
</details>

---

# Changelog

:cl:
-add: Gave BSO access to a filled combat medkit, a portafib, and a handheld crew monitor

### DIFF
--- a/Resources/Prototypes/Loadouts/Jobs/Command/blueshield_officer.yml
+++ b/Resources/Prototypes/Loadouts/Jobs/Command/blueshield_officer.yml
@@ -104,6 +104,17 @@
   items:
     - Portafib
 
+- type: loadout
+  id: LoadoutBSOCrewMonitor
+  category: JobsCommandBlueshieldOfficer
+  cost: 6
+  requirements:
+    - !type:CharacterJobRequirement
+      jobs:
+        - BlueshieldOfficer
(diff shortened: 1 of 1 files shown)


EctoplasmIsGood:
These changes were based on conversations I've had with various players on Goob MRP, and I understand if you'd prefer I move this downstream instead.

VMSolidus (approved):

PR #1796, opened 2025-02-15
Remove Antag Advantage

VMSolidus:
It's kind of cringe.


Aidenkrz (approved):
its really fucking cringe, it sucks so bad. default ee weights for antags are already so low, let the people who do role it be able to properly do their shit

PR #1797, opened 2025-02-15
Damage On High Speed Impact Scales With High Speed Impact

VMSolidus:
# Description

The math used in this system was complete fucking lunacy. Now it runs on a more sane formula that actually allows it to scale with the "High speed impact". The component now operates based on an, "Amount of damage per meters per second of collision", which for a collision that trips it at 10 meters per second, is exactly 5 blunt damage(which is what it was previously for most things that had this). Except now as the speed increases above 10 meters per second, so does the damage.

# Changelog

:cl:
- add: Being thrown into a wall now causes damage that scales with the speed of the collision.


Aidenkrz (approved):
Fuck. Yes.

VMSolidus:
I would not recommend merging this until I'm not in a crashy state at 2 AM in the morning, and can look at it properly.

Content.Shared/Damage/Systems/DamageOnHighSpeedImpactSystem.cs
```diff
@@ -38,16 +40,17 @@ private void HandleCollide(EntityUid uid, DamageOnHighSpeedImpactComponent compo
         if (speed < component.MinimumSpeed)
             return;
 
+        var speedRatio = speed / component.MinimumSpeed;
```
VMSolidus, line 43:
```suggestion
        var speedRatio = Math.Max(speed - component.MinimumSpeed, 0);
```
Damnit "33 hours with no sleep Tyler", you can't just make it multiplicatively scale with speed. What if someone sets the engine speed limit to something absurd like 300 meters per second, and now you get player characters taking a million or so blunt damage? Rewrite this with linear scaling damage.

VMSolidus:
Sending coderabbit to fact check my math here while im in class.

Content.Shared/Damage/Systems/DamageOnHighSpeedImpactSystem.cs
Timfa2112: @VMSolidus was this resolved yet?

EctoplasmIsGood:
@VMSolidus Why was this never merged?

VMSolidus:
> @VMSolidus Why was this never merged?

I decided I didn't like the math at all(I made it while drunk), and wanted to revisit the idea entirely after I got more Physics fundamentals under my belt. I do now, so I'll be revisiting this soon anyways.

PR #1800, opened 2025-02-15
Lambda Station: Minor Fixes

Ichaie:
# Description

Makes minor fixes to Lambda and removes NRP from emergency_raven

---

# Changelog

:cl: Ichai
- tweak: Lambda: added hardsuit for paramedic, added missing chemvend, add missing bounty computer on cargo.
- remove: Remove the NRP (L6 and RPG) from emergency_raven


Ichaie:
I was unable to correct the error with the crew monitor server, but the problem can be temporarily resolved just using wrench

VMSolidus (approved):

PR #1805, opened 2025-02-16
Prevent SecretRule From Picking Invalid Presets (#27456)

sleepyyapril:
trying to fix the prelobby heisentest

we use votes anyway! fuck it, we ball


VMSolidus:
Oh yea fuck the meteor rework. That's the update that was so comically overtuned that it prompted Wizden to make the infamous "Space is so harmless, you can survive it indefinitely by taking a nap in a bed" PR. I actively encourage you to skip over the meteor scheduler update, and probably just yeet the meteor scheduler complaints from the YML Linter.

VMSolidus (approved):

PR #1809, opened 2025-02-16
Make Footstep Entity Better

lzk228:
# Description

Add some components to the footstep entity to make it be more like the puddle entity


Eagle-0:
There's a pr somewhere that gives forensics to footsteps, shoe size and all. Can't quite remember what repo it was on.

Eagle-0:
Ok I lied about the shoe size but it does add forensics. Pretty sure this is where the footsteps originally came from.
https://github.com/Mira-Sector/space-station-14/pull/58

DEATHB4DEFEAT (approved):

PR #1815, opened 2025-02-17
Disable Unecessary Roundstart Variations

VMSolidus:
# Description

EXPLANATION: Mappers need to be trusted to detail their own maps, and not need nonsensical RNG to do their map detailing for them. By far the most egregious sin is the light breaking pass, because it forces mappers to place excessive numbers of lights in order to accomodate for "At least 30% of them will be randomly broken", which means that it is by extension impossible to properly Detail a map for lighting effects, because you can't guarantee that the shadows and lowlights you intend the station to have, will even exist at all. 

The same goes for Rusting passes. Wall rusting is actually intended for demonstrating there's "Zelda Doors", and by adding random rust variation, it is now impossible to hint at the existence of Zelda Doors. If we wanted to "Weather" stations to give them an impression of being used, we would actually want to create mapper assets for walls in variations of rusting, so that they can be more selective about where, when, and how much to use weathering.

The last one being yeeted, bloodstain passes, is for similar detailing reasons. Bloodstains need to be intentionally mapped and not left to RNG because they serve an actual gameplay purpose. Bloodstains and the footprints they inevitably create are a visible sign of a crime happening, an organic form of environmental storytelling. RNG doesn't create a story, it just inapprorpiately places random bloodstains where nothing happened. Which disrupts its usefullness for that purpose. 

<p>

Example: A medbay lit carefully and intentionally, with every individual light having a carefully considered reason for its placement. The room gets to have natural shadows and gradation without being always DROWNED in glow.

image

BAD EXAMPLES:

image

image

image

</p>
</details>

# Changelog

:cl:
- remove: Disabled random variation for Light Breaking, Wall Rusting, and Bloodstains. Our new maps have more carefully measured light usage, and all three of these interfere with good mapping practice.


sleepyyapril:
This PRs my test dummy, re-running tests

OldDanceJacket (approved):

PR #1843, opened 2025-02-25
Holoparasite Buff

EctoplasmIsGood:
# Description

Buffs holoparasite and brings it more in line with its inspiration. Currently, holoparasite attacks very slowly but does high damage per hit, regardless they're very weak and tend to die before they can actually take anybody down. With my buff, it makes them do less damage per hit, but WAY faster, dramatically decreasing their TTK. This buff brings them more in line with the fast-hitting barrages of stands (their inspiration), and also makes them actually viable, and worth their price again.

Also for reference, they cost 10tc more then a damn singularity beacon, and currently kill someone slower then a random guy with a crowbar would.

<p>

</p>
This is a holoparasite CURRENTLY

https://github.com/user-attachments/assets/64f724e8-7d02-4047-bd8b-78d056259b57

This is a holoparasite with my buffs

https://github.com/user-attachments/assets/654de634-2909-4771-96ea-4b06a0135c6f

</details>

---

# Changelog

:cl:
- tweak: Holoparasite melee damage from 20 blunt to 10 blunt
- tweak: Holoparasite melee speed from 1.8 to 0.3

### DIFF
--- a/Resources/Prototypes/Entities/Mobs/Player/guardian.yml
+++ b/Resources/Prototypes/Entities/Mobs/Player/guardian.yml
@@ -138,14 +138,14 @@
       hidden: false
       altDisarm: false
       animation: WeaponArcFist
-      attackRate: 0.35
+      attackRate: 0.3
       autoAttack: true
       soundHit:
         collection: Punch
       damage:
         types:
-          Blunt: 7
-          Structural: 7
+          Blunt: 10
+          Structural: 10
     - type: HTN
       rootTask:
         task: SimpleHumanoidHostileCompound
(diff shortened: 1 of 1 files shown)


EctoplasmIsGood:
Pretty drastic buff considering their TC cost, how easy they are to kill, and the fact that they're an actual ghost role. 

I have no clue why they're so weak right now, they're MEANT to be strong and scary.

EctoplasmIsGood:
also ignore that the media is called "ifrit pre buff" and "ifrit post buff" this is for holoparasites, I just had a brainfart.

VMSolidus (approved):
Oh wack they were never updated.

PR #1848, opened 2025-02-27
Fix Various Psionic Bugs

VMSolidus:
# Description

This isn't fixing all of them by the way, I still need to completely rewrite the equations for the glimmer generating structures so the math there isn't completely terrible. This fixes a few bugs related to the glimmer system, such as linear decay not being consistent regardless of server ping, and also glimmer generating structures also not being consistent based on server ping. Both are now correctly differentiated with respect to time.

Problem though is that the glimmer generator math is still extremely awful, so that's going to need a rewrite.

After deeply analyzing the glimmer equations, I figured out that they were being screwed over by floating point precision limits, so to address this I made Glimmer work on doubles instead. 

I also fixed a bug where generating enough Potentia to gain two or more powers in a single draw caused the user to permanently lose their ability to generate new powers. It's not abundantly clear why this was happening, because nothing changed this code at all since I last touched it. 

# Changelog

:cl:
- fix: Fixed various bugs with psionics.
- fix: Fixed a bug where if you generated enough Potentia during a drug induced psionics roll to gain two powers, you would only gain one power instead, while permanently losing the ability to gain new powers.
- tweak: Glimmer now operates on double precision. Have fun with that breaking change if you made any psionic code downstream. Hint, you can cast doubles into a (float) if you don't wish to convert your system to double.

### DIFF
--- a/Content.Server/Psionics/Glimmer/GlimmerReactiveSystem.cs
+++ b/Content.Server/Psionics/Glimmer/GlimmerReactiveSystem.cs
@@ -178,7 +178,7 @@ public sealed class GlimmerReactiveSystem : EntitySystem
     private void OnDamageChanged(EntityUid uid, SharedGlimmerReactiveComponent component, DamageChangedEvent args)
     {
         if (args.Origin == null
-            || !_random.Prob(_glimmerSystem.GetGlimmerEquilibriumRatio() / 10))
+            || !_random.Prob((float) _glimmerSystem.GetGlimmerEquilibriumRatio() / 10))
             return;
 
         var tier = _glimmerSystem.GetGlimmerTier();
@@ -195,13 +195,14 @@ public sealed class GlimmerReactiveSystem : EntitySystem
         if (tier < GlimmerTier.High)
             return;
 
-        var totalIntensity = _glimmerSystem.GlimmerOutput * 2;
-        var slope = 11 - _glimmerSystem.GlimmerOutput / 100;
+        var glimmer = (float) _glimmerSystem.GlimmerOutput;
+        var totalIntensity = glimmer * 2;
+        var slope = 11 - glimmer / 100;
         var maxIntensity = 20;
 
-        var removed = _glimmerSystem.GlimmerOutput * _random.NextFloat(0.1f, 0.15f);
+        var removed = glimmer * _random.NextFloat(0.1f, 0.15f);
         _glimmerSystem.DeltaGlimmerInput(-removed);
-        BeamRandomNearProber(uid, (int) _glimmerSystem.GlimmerOutput / 350, _glimmerSystem.GlimmerOutput / 50);
+        BeamRandomNearProber(uid, (int) glimmer / 350, glimmer / 50);
         _explosionSystem.QueueExplosion(uid, "Default", totalIntensity, slope, maxIntensity);
     }
 
--- a/Content.Server/Psionics/PsionicsSystem.cs
+++ b/Content.Server/Psionics/PsionicsSystem.cs
@@ -225,7 +225,8 @@ public sealed class PsionicsSystem : EntitySystem
         RollAgain:
         component.Potentia -= component.NextPowerCost;
         _psionicAbilitiesSystem.AddPsionics(uid);
-        component.NextPowerCost = component.BaselinePowerCost * MathF.Pow(2, component.PowerSlotsTaken);
+        component.NextPowerCost = Math.Abs( // Taking the absolute value here as a logical guard against this GOTO infinitely looping.
+            component.BaselinePowerCost * MathF.Pow(2, component.PowerSlotsTaken));
         if (component.Potentia > component.NextPowerCost)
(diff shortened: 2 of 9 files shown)


VMSolidus:
@coderabbitai full review

VMSolidus:
That's odd, where'd the funny naval officer go?

Content.Server/Psionics/PsionicsSystem.cs
```diff
@@ -222,9 +222,13 @@ private bool HandlePotentiaCalculations(EntityUid uid, PsionicComponent componen
...
 
-        component.Potentia -= component.NextPowerCost;
-        _psionicAbilitiesSystem.AddPsionics(uid);
-        component.NextPowerCost = component.BaselinePowerCost * MathF.Pow(2, component.PowerSlotsTaken);
+        while (component.Potentia >= component.NextPowerCost)
+        {
+            component.Potentia -= component.NextPowerCost;
+            _psionicAbilitiesSystem.AddPsionics(uid);
+            component.NextPowerCost = Math.Abs(component.BaselinePowerCost * MathF.Pow(2, component.PowerSlotsTaken));
+        }
```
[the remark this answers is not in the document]
VMSolidus: I'm actually more concerned about some random contributor or downstream setting it to a negative number via the datafield, rather than another system touching it. Since its humans who will ultimately be touching this number rather than a machine, I gotta somehow sanity check it that they can't give it a bad input.

VMSolidus:
Why the hell has nobody reviewed this yet? I'm not doing anything more with this PR, anything further is out of scope, and this has key bugfixes anyways.

Content.Server/Psionics/PsionicsSystem.cs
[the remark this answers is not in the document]
VMSolidus: Changing values of datafields like this during initialization is strictly forbidden for reasons related to the game engine. This game loads maps and then resaves them locally during init, such that if something modifies anything at all in the map during initalization, it will permanently alter the map saved to the server until the next time the server updates. As to why this limitations exists is beyond my current understanding, but we have a test that WILL complain about doing this suggestion.

Also as a general convention, and this isn't a hard requirement but we do enforce it quite intensely, but components are not allowed to contain logic. We have an extremely strict separation of System from Component, where Components are only allowed to contain raw data and nothing else, and systems are explicitly forbidden from containing data. You will probably fact check this and find out that we have a lot of legacy systems that break these rules constantly, as well as new systems that break these rules. That's mostly fine, if there genuinely is no better way to do something while staying in the bounds of ECS, then we'll just do it.

VMSolidus:
image

The simplest proof I can give that the new output equation works correctly.

VMSolidus:
Why do I have 6 unrelated test fails now?

Content.Server/Research/Oracle/OracleSystem.cs
```diff
@@ -178,7 +178,7 @@ private void DispenseLiquidReward(Entity<OracleComponent> oracle)
             return;
 
         // Why is this hardcoded?
-        var amount = MathF.Round(20 + _random.Next(1, 30) + _glimmer.GlimmerOutput / 10f);
+        var amount = MathF.Round(20 + _random.Next(1, 30) + (float) _glimmer.GlimmerOutput / 10f);
```
Timfa2112, line 181: I think there's an argument to be made for making consts out of these numbers--what are these values, and why must they be between 21 and 50?

Content.Server/Psionics/Glimmer/GlimmerReactiveSystem.cs
```diff
@@ -195,13 +195,14 @@ private void OnDestroyed(EntityUid uid, SharedGlimmerReactiveComponent component
...
+        var glimmer = (float) _glimmerSystem.GlimmerOutput;
+        var totalIntensity = glimmer * 2;
+        var slope = 11 - glimmer / 100;
         var maxIntensity = 20;
 
-        var removed = _glimmerSystem.GlimmerOutput * _random.NextFloat(0.1f, 0.15f);
+        var removed = glimmer * _random.NextFloat(0.1f, 0.15f);
         _glimmerSystem.DeltaGlimmerInput(-removed);
-        BeamRandomNearProber(uid, (int) _glimmerSystem.GlimmerOutput / 350, _glimmerSystem.GlimmerOutput / 50);
+        BeamRandomNearProber(uid, (int) glimmer / 350, glimmer / 50);
```
Timfa2112, line 205: Same here--I think it would make sense to make these consts instead of magic numbers in the code.

Timfa2112 (approved):
I don't see much wrong with this, but I do see a few magic numbers that could be made into constants :)

Content.Server/Psionics/Glimmer/GlimmerReactiveSystem.cs
VMSolidus: I'll hit most of these in a separate PR anyways.

PR #1886, opened 2025-03-07
BSO Misc Buffs

WazuraWasTaken:
# Description

Gives BSO accessed to the energy sword for 6 points as a primary weapon: Okay so my thought process for this is that, BSO currently has these options the; BRDI is all around good although expensive, the energy shield that might be replaced by the ones from goob, the Chester... after the .45 ammo changes its never going to be used(like everyone was using it in the first place to begin with) and last and my favorite the energy shotgun which after the it received is now in a fantastic place. Now, why the energy sword? Simple it looks cool and has a niche, which is; tanking and reflecting shots back at enemies and also being a great melee weapon. Have I mention it just looks sick and the intimidation factor of it!
Alright aside from that I think it fits the role of the BSO quite well, since your suppose to run in and be a shield for command. Granted the shield from goob would pretty much fill this role, but can it light a cigarette?

Allow BSO to use Corpsman's survival box: They are basically the upgraded and more well armed version of the corpsman, but for command. I find it strange, that they use the basic survival kit instead of this one.

---

<p>

Example Media Embed

</p>
</details>

---

# Changelog

:cl:
- add: Added Energy Sword to BSO's Arsenal 
- tweak: Allows BSO to take corpsman's survival kit

### DIFF
--- a/Resources/Changelog/Changelog.yml
+++ b/Resources/Changelog/Changelog.yml
@@ -13861,3 +13861,16 @@ Entries:
   id: 7046
   time: '2025-03-27T11:46:02.0000000+00:00'
   url: https://github.com/Simple-Station/Einstein-Engines/pull/2095
+- author: BramvanZijp
+  changes:
+    - type: Fix
+      message: >-
+        IPCs are no longer prevented from raising their shield due to the tile
+        being blocked by their own brain.
+    - type: Fix
+      message: >-
+        If you attempt to raise your shield but get blocked by another mob, you
+        will now only get one popup indicating this.
+  id: 7047
+  time: '2025-03-27T11:47:50.0000000+00:00'
+  url: https://github.com/Simple-Station/Einstein-Engines/pull/2094


EctoplasmIsGood:
no

Resources/Prototypes/Loadouts/Jobs/Command/blueshield_officer.yml
```diff
@@ -67,6 +67,23 @@
...
+  requirements:
+    - !type:CharacterItemGroupRequirement
+      group: LoadoutBlueshieldOfficerPrimary
+    - !type:CharacterJobRequirement
+      jobs:
+        - BlueshieldOfficer
+    - !type:CharacterAgeRequirement
+      min: 21
+  items:
+    - EnergySword
```
VMSolidus, line 85: They actually already have this available from the Security Weapons selection that they already have access to. So this part of the PR isn't strictly necessary.

VMSolidus (changes requested):

Resources/Prototypes/Loadouts/Jobs/Command/blueshield_officer.yml
EctoplasmIsGood: They have access to the weaker energy cutlass, this PR gives them access to the energy SWORD, which is not needed at all and way too powerful to throw into their kit

WazuraWasTaken: I mean this is at the cost of the other primary weapons for bso and it does cost an extra 6 points. Which makes the bso choose between the meds or cqc.

PR #1896, opened 2025-03-08
Throw Errant Into The Spider Pit

VMSolidus:
# Description

@Errant-4 by request in your comment.
I had annihilated this before during the NyanoCombat reworks, but during a recent wizmerge most of said reworks were accidentally reverted. This is the one that was brought back, comment and all. So this PR fixes for a second time a problem left by the old combat system. 

# Changelog

:cl:
- fix: Power attack and light attack damages are no longer swapped.


Erisfiregamer1:
image
We don't *have* a spider pit, though... Poor man's just getting thrown in the Supermatter pit instead

angelofallars (approved):

PR #1906, opened 2025-03-09
Trait Cost Balancing

EctoplasmIsGood:
# Description

Makes a lot of disabilities give more points (based off community and personal opinions), and makes some very strong traits cost more points.

# Changelog

:cl:
- tweak: Narcolepsy gives 6 points instead of 2
- tweak: Pacifist gives 8 points instead of 6
- tweak: Muted gives 6 points instead of 4
- tweak: Sluggish gives 5 points instead of 3
- tweak: Snail Paced gives 8 points instead of 5
- tweak: Blood Deficiency gives 6 points instead of 5
- tweak: Hemophilia gives 4 points instead of 2
- tweak: Clumsy gives 4 points instead of 1
- tweak: Bad knees gives 5 points instead of 3
- tweak: Will to live costs 3 points instead of 1
- tweak: Tenacity costs 4 points instead of 3
- tweak: Steadfast costs 6 points instead of 4
- tweak: Bionic Arm costs 10 points instead of 8
- tweak: Bionic leg costs 8 points instead of 6

### DIFF
--- a/Resources/Prototypes/Traits/disabilities.yml
+++ b/Resources/Prototypes/Traits/disabilities.yml
@@ -57,7 +57,7 @@
 - type: trait
   id: Narcolepsy
   category: Mental
-  points: 2
+  points: 6
   requirements:
     - !type:CharacterJobRequirement
       inverted: true
@@ -78,7 +78,7 @@
 - type: trait
   id: Pacifist
   category: Mental
-  points: 6
+  points: 8
   functions:
     - !type:TraitAddComponent
       components:
@@ -101,7 +101,7 @@
 - type: trait
   id: Muted
   category: Mental
-  points: 4
+  points: 6
   requirements:
     - !type:CharacterJobRequirement
       inverted: true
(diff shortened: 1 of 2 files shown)


Raikyr0:
We will argue forever over the Bionic Arm, but I think it's in a perfectly fine spot for the point value as we both know the other end of the points spectrum is, halving your HP.

I would say Redshirt should be bumped to 10, as if they die to radiation or poison, they're a real pain to recover.
While Osteogenesis Imperfecta could even be bumped to 12. As this trait, results in a near immediate death in any combat.
An increase to Blood Deficiency, would also be nice as it's rather difficult to source Saline or Blood Packs for long term.
Note that Sluggish, makes you slower then Bionic Leg's speed boost would provide in the other direction.
Feeble should cost the same as Steadfast as it does the same thing in the other direction.

Ultimately I find the issue isn't cost, but the max number of traits you can take. We may chrome max all we want, but I think psionics should be addressed rather then making cybernetics more expensive.

VMSolidus (approved):

PR #1880, opened 2025-03-05
NT Poster Censorship

EctoplasmIsGood:
# Description

Removes multiple posters that shouldn't be ingame.

# Changelog

:cl:
- remove: You heard nothing and you saw nothing, glory to Nanotrasen.

## Summary by CodeRabbit

- **New Features**
  - Added new entries for poster entities to enhance management and representation within the system.

- **Chores**
  - Commented out specific poster entries from random spawning logic for a more curated display.
  - Minor formatting adjustment in the YAML file by removing an unnecessary blank line.

PR #1919, opened 2025-03-10
Revert "NT Poster Censorship" (#1880)

BasedUser:
# Description

This PR re-adds the Enlist, and EnlistGorlex posters. Why:
- The Enlist poster has been replaced from a Death Squad operative to ERT ***ages ago*** in Delta-V code. Regardless of whatever the final codebase using this poster, I don't think this warrants a complete migration removal, since ERT are a known substance almost universally, and the closest example, like the Sec ERT, looks as distinct from a DS as both from a blood-red.
- The EnlistGorlex poster doesn't actually make mention of any nuclear operatives. It is a task Gorlex sends agents on, yes, but from a PR point of view blatantly advertising nuclear capabilities would ruin the balance of power.
  "Oh, but the hardsuit! That's a nukie hardsuit!" Wrong.
  
  Traitors - even non-Gorlex traitors - can buy those hardsuits. The hardsuit alone wouldn't imply nuclear operatives exist. It's scary, but not THAT scary, even as a contraband poster.

In absolutely no circumstance would I put them in migrations unless the lore SPECIFICALLY removes both nuclear operatives (which are still in fact a gamemode) and ERTs (which are still an IC superforce). I can see a situation where you'd want an Enlist poster upstream but not in the marker spawn list, for example.

---

# TODO

This PR is complete, I do not need help completing it.  

---

# Changelog

:cl: router
- add: Re-added ERT and Gorlex enlist posters to the default roster, and removed them from the mapping migration list.
- remove: Removed No ERP posters from rotation. This does not affect specifically placed posters.

### DIFF
--- a/Resources/Migrations/migration.yml
+++ b/Resources/Migrations/migration.yml
@@ -316,9 +316,6 @@ ImprovisedExplosiveFuel: FireBombFuel
 WeaponSubMachineGunVectorRubber: null
 WeaponSubMachineGunVector: null
 
-# 2025-03-10
-PosterLegitNoERP: null
-
 # Lavaland EE stuff
 TreasureDatadiskEncrypted: null
 VendingMachineBoozeSyndicate: VendingMachineBooze
--- a/Resources/Migrations/migration.yml
+++ b/Resources/Migrations/migration.yml
@@ -316,6 +316,9 @@ ImprovisedExplosiveFuel: FireBombFuel
 WeaponSubMachineGunVectorRubber: null
 WeaponSubMachineGunVector: null
 
+# 2025-03-10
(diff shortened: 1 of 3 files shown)


Eagle-0:
I agree, no clue why these were removed in the first place

EctoplasmIsGood:
I don’t care about these being reverted, especially because DS is receiving a rework to make them even more independent from ERT.

VMSolidus:
image-179

Observe as your spite revert gets closed.

Farida Sing 1k transparent

EctoplasmIsGood:
Crazy how that works

BasedUser:
lmao

PR #1923, opened 2025-03-10
Fairer Turrets

VMSolidus:
# Description

It had basically always bothered me, and evidently a lot of other people that turrets were previously "Unfair enemies". They had perfect aimbot accuracy, tracked targets instantly at any angle, and were nigh-indestructible. It was telling that players were pretty much only ever killed by turrets once or twice, and then they figure out that the turrets are outright impossible to beat in a fair fight, so they end up resorting to underhanded tactics like using an RCD to wall the turret off. Or use a pickaxe to go around them.

I previously added a new turret for the SAN dropship, and with it had tried out a very different approach for "Fairer Turret" balancing. A more reasonable healthbar, no instant tracking, and no perfect accuracy, The players enjoyed these a lot more than the original turrets. Plus they were more visually interesting firing an inaccurate spray at their targets.

<p>

I'm on my lunch break at college right now, so no media.

</p>
</details>

# Changelog

:cl:
- tweak: Adjusted turrets to be a more "Fair" fight. They no longer have perfect aimbot accuracy, they don't track targets instantly, and they also have a "More fair" healthbar that can reasonably be shot through by someone with good enough weapons and armor,

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Weapons/Guns/turrets.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Guns/turrets.yml
@@ -73,51 +73,80 @@
       thresholds:
         - trigger:
             !type:DamageTrigger
-            damage: 600
+            damage: 200 # Was 600, most players were under the impression these were indestructible. It having an Inorganic damage container did not help.
           behaviors:
             - !type:DoActsBehavior
               acts: [ "Destruction" ]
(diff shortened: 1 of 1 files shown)


SolarisBirb:
Based. Currently, the turrets are super strong. If you walk into any room with turrets you're good as dead.

angelofallars (approved):

PR #1947, opened 2025-03-13
Fix Cult Doors

VMSolidus:
# Description

Cult doors would launch people at FTL speed, they exerted 13,000 Newtons of force on non-cultists who touched them. Which while sometimes hilarious, is enough to send people through walls. This PR sets it to a more reasonable number of 300 Newtons, which is enough to throw a standard human about 2 tiles from the door.

# Changelog

:cl:
- tweak: Tweaked cult doors so that they don't launch people at luminal speeds.


SolarisBirb:
LOL I didn't even know this was a bug. That's so funny.

EctoplasmIsGood:
2 tiles isnt enough, should be like 10

OldDanceJacket (approved):

PR #1948, opened 2025-03-13
Lavaland Shuttle Fix

SolarisBirb:
# Description

The bug in question is the one where if you return from Lavaland the shuttle would not dock to the station but instead send you approximately around the station.

I replaced the original function and lookup completely where when the station and lavaland are initialized it will add them to destinations and then use a list to warp to the gridUIDs. This allows for more consistent warping and less time for looking up what the largest grid is. This fix also fixes the issue of smaller maps like Northway where a wreck could be much larger than the station and would not warp to it.

Here is the new format for the mining shuttle console
image
---

# TODO

- [x] Rework the shuttle, adding destinations to make it more modular. 

---

# Changelog

:cl:
- fix: NT has finally patched the mining shuttle and now docks to the station on returning from lavaland.


Erisfiregamer1 (approved):
I salute you for fixing this in record time.

SolarisBirb:
> I genuinely do not like this code at all, and it's actually still more hardcoded than I'd like. It should probably work based on a FTLTargetComponent that you put on the airlocks, and have the shuttle work by enumerating the list of Entity to find the possible docks it can travel to. It's actually trivial to find the grid that an FTLTargetComponent is on, because the Entity it belongs to will have a TransformComponent which declares what its gridUid and mapUid are.
> 
> Did you know that this might actually break calling the shuttle to lavaland, because Lavaland Outpost won't have a StationDataComponent since it isn't a place where you spawn jobs? It also won't work on any arbitrary grid with a console on it if it doesn't have this component.

I can work on a better solution. I know that it sounds strange but the reason why I went for this solution is due to how they coded the outpost, it was set as both a StationDataComponent and also a LavalandStationComponent. It seems to have both components. I just know one thing for sure the find largest grid function did not work at all. I feel more of a list with gridUIDs when selecting locations would work better. Are we planning on adding warp points that are not station->lavaland?

VMSolidus:
> > I genuinely do not like this code at all, and it's actually still more hardcoded than I'd like. It should probably work based on a FTLTargetComponent that you put on the airlocks, and have the shuttle work by enumerating the list of Entity to find the possible docks it can travel to. It's actually trivial to find the grid that an FTLTargetComponent is on, because the Entity it belongs to will have a TransformComponent which declares what its gridUid and mapUid are.
> > Did you know that this might actually break calling the shuttle to lavaland, because Lavaland Outpost won't have a StationDataComponent since it isn't a place where you spawn jobs? It also won't work on any arbitrary grid with a console on it if it doesn't have this component.
> 
> I can work on a better solution. I know that it sounds strange but the reason why I went for this solution is due to how they coded the outpost, it was set as both a StationDataComponent and also a LavalandStationComponent. It seems to have both components. I just know one thing for sure the find largest grid function did not work at all. I feel more of a list with gridUIDs when selecting locations would work better. Are we planning on adding warp points that are not station->lavaland?

Oh totally, per discord conversations. The console supports having a list of destinations, but is currently hardcoded around only ever having two destinations in that list. I would LOVE to have something like a "Public transit shuttle" that can be setup to warp between any number of destinations, so that we can do something like the Frontier style of map with multiple stations.

SolarisBirb:
> > > I genuinely do not like this code at all, and it's actually still more hardcoded than I'd like. It should probably work based on a FTLTargetComponent that you put on the airlocks, and have the shuttle work by enumerating the list of Entity to find the possible docks it can travel to. It's actually trivial to find the grid that an FTLTargetComponent is on, because the Entity it belongs to will have a TransformComponent which declares what its gridUid and mapUid are.
> > > Did you know that this might actually break calling the shuttle to lavaland, because Lavaland Outpost won't have a StationDataComponent since it isn't a place where you spawn jobs? It also won't work on any arbitrary grid with a console on it if it doesn't have this component.
> > 
> > 
> > I can work on a better solution. I know that it sounds strange but the reason why I went for this solution is due to how they coded the outpost, it was set as both a StationDataComponent and also a LavalandStationComponent. It seems to have both components. I just know one thing for sure the find largest grid function did not work at all. I feel more of a list with gridUIDs when selecting locations would work better. Are we planning on adding warp points that are not station->lavaland?
> 
> Oh totally, per discord conversations. The console supports having a list of destinations, but is currently hardcoded around only ever having two destinations in that list. I would LOVE to have something like a "Public transit shuttle" that can be setup to warp between any number of destinations, so that we can do something like the Frontier style of map with multiple stations.

You got it, boss. I will clean up my shit code and get this working later tonight! :)

Content.Server/_Lavaland/Shuttles/Systems/DockingShuttleSystem.cs
SolarisBirb, line 1: AddStation function needs to be changed to check for stations not station datacomponent. This causes issues since if you run the function ATS, cargo shuttle, pathfinder, and other shuttles have this component.

SolarisBirb: Fixed it. Now it will pull only stations and not use the stationdatacomponent.

Content.Server/_Lavaland/Shuttles/Systems/DockingShuttleSystem.cs
```diff
@@ -83,33 +88,135 @@ private void OnStationGridAdded(StationGridAddedEvent args)
...
+
+    /// <summary>
+    /// This function will specifically for lavaland components or station components on any given map.
+    /// This will allow for you to add more maps or have many stations/lavaland structures to warp too :)
+    /// </summary>
+    private void AddDestinations(DockingShuttleComponent component, MapId map, EntityUid uid)
+    {
+        var planet = Name(uid).Split();
+        // TODO: Change this to a switch/case if there is more station based components
+        if (planet[0] != "Lavaland")
```
VMSolidus, line 141: I don't actually think this should be hardcoded like this, when realistically this kinda isn't necessary to differentiate "Station vs. Lavaland"? Lavaland's station can totally just have its name changed to something that reflects its status as a Lavaland mining colony. And have the destinations just use the name of the station as a target. So if Lavaland had the same kinda naming conventions as Radstation here, then including the "Station - StationName" OR "Lavaland - StationName" part is unecessary. It can just be "StationName" as the destination name, without a need to differentiate the two.

image

VMSolidus (changes requested):

Content.Server/_Lavaland/Shuttles/Systems/DockingShuttleSystem.cs
VMSolidus: Example: "NT14 Lavaland Mining Outpost" as the destination, and that can just be the name of the lavaland grid that it docks to. :)

SolarisBirb: Alrighty made adjustments for you. Let me know if there is anything else you would like me to change!

VMSolidus (approved):

PR #1949, opened 2025-03-13
Child of the Noosphere

dootythefrooty:
# Description

Adds the "Scarier Mindbreaking" which already existed as a CCvar, as an optional trait. You're effectively round removed if you get mindbroken as a latent with this trait. 

# Changelog

:cl:
- add:  Child of the Noosphere", a psionic trait that increases the effect of mindbreaking to an extreme, has been added.

### DIFF
--- a/Content.Server/Abilities/Psionics/PsionicAbilitiesSystem.cs
+++ b/Content.Server/Abilities/Psionics/PsionicAbilitiesSystem.cs
@@ -15,6 +15,8 @@ using Robust.Shared.Serialization.Manager;
 using Robust.Shared.Player;
 using Content.Shared.CCVar;
 using Content.Shared.NPC.Systems;
+using Content.Shared.Tag;
+
 
 namespace Content.Server.Abilities.Psionics;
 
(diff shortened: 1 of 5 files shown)


RocketBoss360:
should this be available for shadowkin too? (innate psionics, disabled if u got blackeye ofc)

Content.Server/Abilities/Psionics/PsionicAbilitiesSystem.cs
```diff
@@ -205,7 +208,7 @@ public void MindBreak(EntityUid uid)
         var ev = new OnMindbreakEvent();
         RaiseLocalEvent(uid, ref ev);
 
-        if (_config.GetCVar(CCVars.ScarierMindbreaking))
+        if (_config.GetCVar(CCVars.ScarierMindbreaking) || _tag.HasTag(uid,"TraitScaryMindbreaking" ))
```
VMSolidus, line 211: This really shouldn't be hardcoded to to do this. You should instead modify the OnMindbreakEvent to include a bool for ForceScarrierMindbreaking, which can be set by any component listening to the OnMindBreakEvent. Then you'll want to add a component for this mindbreaking vulnerability, which has a subscription to ScarierMindBreaking, and when it receives the event, sets said boolean to force it.

VMSolidus (changes requested):

Timfa2112:
@dootythefrooty are you still working on this?

PR #1955, opened 2025-03-13
Remove Random Stencil Clear On Left Leg

Erisfiregamer1:
# Description

I don't know why this was being done, but it didn't seem to do anything and I know for a fact it was causing issues with shaders that were applied on all humanoid layers, so... "Imaginary Technique: Shitcode Removal" go.

If anything breaks, it's this.

---

# TODO

- [X] Check that the game didn't incinerate itself.

---

<p>

image

</p>
</details>

---

# Changelog

no cl


VMSolidus (approved):
I mean you can also delete it instead of commenting. I can confirm its vestigial.

Erisfiregamer1:
> I mean you can also delete it instead of commenting. I can confirm its vestigial.

This is just in case something does somehow catastrophically break, it's an easier find-and-replace to readd it then.

PR #1970, opened 2025-03-15
System to Add Components via Implanters

hobolyra:
# Description

Code handling for adding components to someone via implant. Takes care of all the lifting so that people can use just YML to make said items. Mostly for adding more customization and modding ease of use. 

No current implimentation besides a sample implant to show the YML format, as it was made for another server use, but it might be usefull to others.

Example implant adds the "dispellable" component when implanted, just to show syntax for YML.
```
  - type: AddComponentsImplant
    componentsToAdd:
    - type: Dispellable
```

# Changelog

:cl:
- add: Code files for implanting, removing, and checking for components as stated in YML
- fix: Fixed naming from old server


VMSolidus (approved):
God I hate implant code, but this is a start I guess.

Issue #1956, opened 2025-03-13
Sol Alliance Mod Suit Bundle Issues


Raikyr0:
# Description

Sol Alliance Modsuit has Issues

Power drains too fast when you seal it.
Modsuit Chestpiece appears to be lacking cold resistance.

# Expected behavior

It shouldn't drain 25 of the power just sealing it and start ticking down fast enough that it won't last 5 minutes.

Also the chestpiece provides no resistance to cold temps (I was freezing at 1 degrees)

# Additional context

I'm the only person who buys the Modsuit Bundle

PR #1989, opened 2025-03-17
Power Cell Draw Fixes

VMSolidus:
# Description

Power cell draw wasn't tick invariant, so batteries on this codebase were being drained anywhere from 30 to 60x faster than people expect, depending on server performance. This was notably something that severely effected MODSuits, which can now enjoy a battery lifespan of around 30 minutes. But it's also a big deal for IPCs.

Closes https://github.com/Simple-Station/Einstein-Engines/issues/1956

# Changelog

:cl:
- fix: Batteries no longer drain power 30 to 60 times faster than physics states. This is particularly a large buff to IPCs and MODSuits, who both no longer have to deal with batteries lasting an absurdly short amount of time. 
- tweak: Solarian Modsuit now provides protection from cold.

### DIFF
--- a/Content.Server/PowerCell/PowerCellSystem.Draw.cs
+++ b/Content.Server/PowerCell/PowerCellSystem.Draw.cs
@@ -28,7 +28,8 @@ public sealed partial class PowerCellSystem
             if (!TryGetBatteryFromSlot(uid, out var batteryEnt, out var battery, slot))
                 continue;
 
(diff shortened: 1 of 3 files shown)


Raikyr0:
> batteries on this codebase were being drained anywhere from 30 to 60x faster

holy moly

DEATHB4DEFEAT (approved):

Issue #2001, opened 2025-03-18
Bonfire/Fireplace Recipes Missing


Raikyr0:
# Description

Bonfire/Fireplace Recipes Missing

# Expected behavior

Should be there

PR #2003, opened 2025-03-18
Bonfire Fix

Raikyr0:
# Description

#2001 Attempted Fix, The original code was for Wizden so there was probably a tag or function that was named differently. Changed from DestroyEntity to DeleteEntity. Not sure if this will work tho.

# Changelog

- fix: Fixed Bonfires and Fireplaces being not buildable maybe


Timfa2112:
You mention you're not sure if this will work, which implies that you haven't tested it :P
Can you try running the game locally to see if your fix works?

Raikyr0:
> You mention you're not sure if this will work, which implies that you haven't tested it :P Can you try running the game locally to see if your fix works?

getting it tested now

PR #2013, opened 2025-03-19
Prevent IPCs From Having Their Access Panel Emagged Permanently.

BramvanZijp:
# Description

I get it, IPCs right now are already overtuned, but there are also SEVERAL ways to instantly disable them, which I dont know how to feel about. EMPs I will think on if that should be changed, and if so, how, later on. But I feel like the fact that it pretty much kills them, as they will remain unable to do anything without outside help, without even actually killing them (Meaning you can't spectate unless you ghost out, which round removes you.) is not ideal, since staring at an inverse black circle until you get lucky and somebody who actually knows how IPCs work, and how to help them if they get killed or emped, is not very fun gameplay.

However, the more pressing issue is the emag. If you use it on a locked IPC, it unlocks their panel instantly with zero warning to the IPC, which then allows you to rip out their cell, which has all the same consequences and thus problems as being EMPed. 
However, the emag is actually worse, as upon use, it DELETES the lock component from that IPC, meaning their panel will remain open forever unless an admin re-adds the component, meaning from there on out for the rest of the round, ANYONE can right click you, eject your cell, and you can do NOTHING about it. So this PR is removing that emag interaction entirely, as again, you already have EMPs if you want to instantly render an IPC defenseless with 0 counterplay, but atleast being hit by an EMP doesnt permanently give anyone the ability to shut you off instantly with no way to revert or repair that.

---

# Changelog

:cl: BramvanZijp
- remove: An IPC's maintenancepanel can no longer be emagged to instantly and permanently open it.

### DIFF
--- a/Resources/Prototypes/Entities/Mobs/Player/ipc.yml
+++ b/Resources/Prototypes/Entities/Mobs/Player/ipc.yml
@@ -28,7 +28,7 @@
     # Por algum motivo esse componente é bem bugado, então o "LockTime" é quem diz o tempo de tudo enquanto o unlock só está servindo como um bool (???)
     lockTime: 5
     unlockTime: 5
-    breakOnEmag: false # Emagging an IPC panel is irreversible without admin intervention, and I have a funny feeling making somebody permanently pretty much combat ineffective as their cell can instantly be yanked out, with no do-after bar, counterplay, or even indicator to the victim, is fair.
+    breakOnEmag: false # Emagging an IPC panel is irreversible without admin intervention, and I have a funny feeling making somebody permanently pretty much combat ineffective as their cell can instantly be yanked out, with no do-after bar, counterplay, or even indicator to the victim, is not exactly fair.
   - type: NpcFactionMember
     factions:
       - NanoTrasen
--- a/Resources/Prototypes/Entities/Mobs/Player/ipc.yml
+++ b/Resources/Prototypes/Entities/Mobs/Player/ipc.yml
@@ -28,6 +28,7 @@
     # Por algum motivo esse componente é bem bugado, então o "LockTime" é quem diz o tempo de tudo enquanto o unlock só está servindo como um bool (???)
     lockTime: 5
     unlockTime: 5
+    breakOnEmag: false # Emagging an IPC panel is irreversible without admin intervention, and I have a funny feeling making somebody permanently pretty much combat ineffective as their cell can instantly be yanked out, with no do-after bar, counterplay, or even indicator to the victim, is fair.
   - type: NpcFactionMember
     factions:
       - NanoTrasen


Timfa2112 (changes requested):
It would be nice to be able to restore the lock. However, removing features and not putting back something in their place will only reduce the content of the game.

Instead of this, might it not be easier to create a new PanelLock entity that you can use on an IPC to re-add the component to the IPC? That way they can do something to restore the lock, instead of the interaction just being gone outright. The object could be produced from the Exo-Suit Fabricator for something like a bit of circuitry and a sheet of steel.

Adding a component to an entity is relatively straightforward, so a simple event listener that listens to the interaction and then adds the component if one doesn't exist already shouldn't be hard to achieve.

Timfa2112:
@coderabbitai Can you produce an example for a PanelLockRepairSystem that handles the interaction of a repair item onto an IPC, which will, if it's not there already, add the Panel lock component to the IPC, being consumed in the process?

Assume a PanelLockRepairComponent exists and is added to an entity already. Only worry about the System.

As for @BramvanZijp -- even if the output of the AI is correct, don't copy-paste it directly. We'd rather have the code in the project written by human hands. You never know when the bot might be compromised by the Sol Alliance after all.

BramvanZijp:
> It would be nice to be able to restore the lock. However, removing features and not putting back something in their place will only reduce the content of the game.
> 
> Instead of this, might it not be easier to create a new PanelLock entity that you can use on an IPC to re-add the component to the IPC? That way they can do something to restore the lock, instead of the interaction just being gone outright. The object could be produced from the Exo-Suit Fabricator for something like a bit of circuitry and a sheet of steel.
> 
> Adding a component to an entity is relatively straightforward, so a simple event listener that listens to the interaction and then adds the component if one doesn't exist already shouldn't be hard to achieve.

Be that as it may, I think the ability to instantly snap open an IPCs maintenence panel just is not fair. 
The panel isnt ID/Access locked, which is what the emag is supposed to be used against, but is moreso there to serve as a do-after for opening an IPCs panel.
While yes, again, IPCs are on the stronger side, I feel like people are missing the amount of EXTREME counters to IPCs that exist.
They take DOUBLE damage from heat, which is no minor thing.
If they get EMPed, they are instantly defenseless, with an EMP implant having little to no counterplay beyond recharging power cells.
And if their panel gets opened, you can instantly remove their cell, again, no counterplay.

It being unable to be repaired is one thing, but I just dont think it should be possible at all, since this is a tool with theoretically unlimited uses can render an IPC defenseless in combat with very little you can do about it as the IPC.

Timfa2112:
After a discussion with Bram and others in Discord, we came to the following alternative;

> The IPC gets a locked panel, that only the IPC can open. When unlocked, anyone opening it will require a screwdriver with a doafter to do so.
> If an emag is used, then anyone can unlock the panel, but it still needs a screwdriver to open.

I agree with the abovementioned, personally. However, I disagree with merging this PR as-is (remove emag on IPC) as a temporary fix, my reasoning being that I don't want to remove a feature that's currently working as-designed for an alternative promised to come in the future.

BramvanZijp:
Closing this PR in favour of a 2 step PR.
PR 1 will be making it so the IPC panel is screwdriver based.
PR 2 will be adding a lock that only the IPC can toggle, but can be destroyed via emag.

PR #2014, opened 2025-03-19
Rework the IPC Maintenance Panel.

BramvanZijp:
# Description

IPC panels now work simularly to borg panels, as follows:
You can unlock/lock them instantly if you have the requires access.
Unlike borgs, IPCs start unlockable by anyone, but this can be changed with an access config.
IPCs can ALWAYS unlock/lock their own panel, regardless of the access requirement.
After it is unlocked, you can then use a screwdriver to ACTUALLY open/close the panel.

---

<p>

[Link to a video in contribution-general (file too big to attach)](https://discord.com/channels/1218698320155906090/1218698321053356060/1351936832434077787)

</p>
</details>

---

# Changelog

:cl: BramvanZijp
- add: You can now add an access requirement to an IPC's maintenance panel using an access configurator, though an IPC can always unlock their own panel.
- tweak: After unlocking an IPC maintenance panel, which is now instant, you need to use a screwdriver to actually open it.

### DIFF
--- a/Content.Shared/Access/Systems/AccessReaderSystem.cs
+++ b/Content.Shared/Access/Systems/AccessReaderSystem.cs
@@ -100,6 +100,9 @@ public sealed class AccessReaderSystem : EntitySystem
         if (!reader.Enabled)
             return true;
 
+        if (reader.OwnerHasAccess && reader.Owner == user)
+            return true;
+
         var accessSources = FindPotentialAccessItems(user);
         var access = FindAccessTags(user, accessSources);
         FindStationRecordKeys(user, out var stationKeys, accessSources);
--- a/Content.Shared/Access/Systems/AccessReaderSystem.cs
+++ b/Content.Shared/Access/Systems/AccessReaderSystem.cs
@@ -100,8 +100,11 @@ public sealed class AccessReaderSystem : EntitySystem
         if (!reader.Enabled)
             return true;
 
-        if (reader.OwnerHasAccess && reader.Owner == user)
+        if (reader.OwnerHasAccess && target == user)
+        {
+            LogAccess((target, reader), user);
             return true;
+        }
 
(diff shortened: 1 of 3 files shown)


Content.Shared/Access/Systems/AccessReaderSystem.cs
```diff
@@ -100,6 +100,9 @@ public bool IsAllowed(EntityUid user, EntityUid target, AccessReaderComponent? r
         if (!reader.Enabled)
             return true;
 
+        if (reader.OwnerHasAccess && reader.Owner == user)
+            return true;
```
Pspritechologist, line 104: This seems to be using `Component.Owner` which is deprecated. We shouldn't be adding new code using it.

Pspritechologist (changes requested):

Pspritechologist:
The changelog seems to be a little bungled

Content.Shared/Access/Systems/AccessReaderSystem.cs
Timfa2112: It's decrepated but still being used all over the place, incl. in many new PRs.
Do you happen to know how you're meant to get the entity uid now?

Timfa2112 (approved):
Approved, just waiting for Psprite for one comment—I still use Owner as well, so I'm curious what the new better way to get that EntityUID is too

BramvanZijp:
> The changelog seems to be a little bungled

Changelog updated

Content.Shared/Access/Systems/AccessReaderSystem.cs
BramvanZijp: Working on it.

BramvanZijp: Resolved.

Pspritechologist (approved):

PR #2015, opened 2025-03-19
HOS Samurai Drip

trashalice:
adds the samurai armour from the syndie bundle to hos loadout but repainted, haven't added to loadouts, up to the council if they think its fitting, i just use EE to test because my server doesn't work for test enviro :Godo:

# Description

adds the samurai armour from the syndie bundle to to the game but repainted, haven't added to loadouts, up to the council if they think its fitting, i just use EE to test because my server doesn't work for test enviro :Godo:
---

# Changelog

:cl:
- add: HoS Samurai armour to go with the Antque sheath (not in loadouts, art people can decide if it goes in)


VMSolidus (approved):
Ooh made by Hyenh. Neat.

PR #2018, opened 2025-03-19
Multilingual

Raikyr0:
# Description

Changed the limit to languages you can take from 1 to 3, and the cost of most basic languages to 0.

Why? It's feasible for someone to learn several languages. Also because it still takes up trait slots, this is still balanced.

# Changelog

:cl:

- tweak: Increased the max number of Languages you can take from 1 to 3
- tweak: Basic Languages are now free
- tweak: Sign Language and advanced Tarajan is now -1 from -2

### DIFF
--- a/Resources/Prototypes/CharacterItemGroups/Generic/languageGroups.yml
+++ b/Resources/Prototypes/CharacterItemGroups/Generic/languageGroups.yml
@@ -1,6 +1,6 @@
 - type: characterItemGroup
   id: TraitsLanguagesBasic
-  maxItems: 1
+  maxItems: 3
   items:
     - type: trait
       id: SignLanguage
--- a/Resources/Prototypes/Traits/languages.yml
+++ b/Resources/Prototypes/Traits/languages.yml
@@ -1,7 +1,7 @@
 - type: trait
   id: SignLanguage
   category: TraitsSpeechLanguages
-  points: -2
+  points: -1
   requirements:
     - !type:CharacterItemGroupRequirement
       group: TraitsLanguagesBasic
--- a/Resources/Prototypes/Traits/languages.yml
+++ b/Resources/Prototypes/Traits/languages.yml
@@ -83,7 +83,7 @@
 - type: trait
   id: ValyrianStandard
   category: TraitsSpeechLanguages
-  points: -1
+  points: 0
   requirements:
     - !type:CharacterItemGroupRequirement
       group: TraitsLanguagesBasic
@@ -111,7 +111,7 @@
 - type: trait
   id: Azaziba
   category: TraitsSpeechLanguages
-  points: -1
+  points: 0
   requirements:
     - !type:CharacterSpeciesRequirement
       species:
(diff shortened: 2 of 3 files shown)


Timfa2112 (changes requested):
I've been talking with @OldDanceJacket about this one, and we've had a thought;

The PR as presented doesn't feel 100% there, but that is mostly due to the point cost reduction. The two of us feel that there should not need to be a limit on the number of languages you can pick at all, so long as you have the points and slots for them.

Do you agree with this? If so, you could remove the limit altogether, but revert the change in point cost. (Or, rather, the ones that became 0 cost. Changing costs is OK, but they shouldn't be free)

Raikyr0:
>Do you agree with this? If so, you could remove the limit altogether, but revert the change in point cost. (Or, rather, the ones that became 0 cost. Changing costs is OK, but they shouldn't be free)

I don't really consider them free if they take up a trait slot, that's my take on it atleast. As for the limit, I don't mind removing it, if someone wishes to be silly language maxxing guy.

Resources/Prototypes/CharacterItemGroups/Generic/languageGroups.yml
```diff
@@ -1,6 +1,6 @@
 - type: characterItemGroup
   id: TraitsLanguagesBasic
-  maxItems: 1
+  maxItems: 3
   items:
```
Timfa2112, line 4:
```suggestion
  items:
```

Resources/Prototypes/Traits/languages.yml
```diff
@@ -83,7 +83,7 @@
 - type: trait
   id: ValyrianStandard
   category: TraitsSpeechLanguages
-  points: -1
+  points: 0
```
Timfa2112, line 86:
```suggestion
  points: -1
```

Resources/Prototypes/Traits/languages.yml
```diff
@@ -111,7 +111,7 @@
 - type: trait
   id: Azaziba
   category: TraitsSpeechLanguages
-  points: -1
+  points: 0
```
Timfa2112, line 114:
```suggestion
  points: -1
```

Resources/Prototypes/_EE/Traits/languages.yml
```diff
@@ -1,7 +1,7 @@
 - type: trait
   id: SiikMaas
   category: TraitsSpeechLanguages
-  points: -1
+  points: 0
```
Timfa2112, line 4:
```suggestion
  points: -1
```

Timfa2112 (changes requested):
Leaving this suggestion here, which is the current recommendation from ODJ
Feel free to discuss with @OldDanceJacket though--I'll go with his recommendation if the two of you come up with an alternative plan :)

Raikyr0:
I have never ever, seen someone use any of the extra Tarajan languages,

The only language anyone ever takes extra is Sign. I was considering adding a Foreigner with which doesn't start with a translator.

Timfa2112:
You don't need to keep updating the branch, even if it's outdated it'll be able to merge just fine.
Only really necessary if there's things in Master you need, or if you have merge conflicts

PR #2028, opened 2025-03-20
Throwing Nerfs: Less Stamina Damage, Cooldown After Grabbing Before You Can Throw.

BramvanZijp:
# Description

This PR nerfs the ability to stun somebody by throwing them after hard-grabbing them, it does this in 2 ways:
1. Stamina damage goes from 120 to 100, Which I am told by eagle will knockdown as opposed to paralyzing.
2. The big one, after you INITIATE your grab on somebody, a 6 second cooldown starts before you can throw them. This cooldown does not reset if you upgrade your grab, and starts as soon as you soft-grab somebody (or instantly hard-grab, for martial artists.)

This is done because right now melee fights involving martial artists are quite bland and simple: Whoever can get the opponent in a hard grab and throw them first, wins, because this instantly stamina crits them, allowing you to cuff them. This completely negates the unique combos/differences between the fighting styles, as this works for Judo, CQC, and SCarp. You grab em, you toss em, you restrain em, you win. And I dont think fights like these should be decided in the first 1-2 seconds based on who reacted SLIGHTLY faster and/or who had better ping, while they could be so much cooler and more in-depth.

But now, if you are in a hard-grab, you have time to try to get loose, be that via resisting, attacking your grabber, or your own Martial Arts moves, BEFORE they can throw and stam-crit you. Furthermore, this should not be that much of a nerf to those who are NOT martial arts trained, as the timer starts upon INITIATING the grab, meaning if a basic tider soft grabs you, and then upgrades to a hard grab, they can throw you after 6 seconds have passed since that initial soft-grab. But if the BSO with CQC instantly hard-grabs you, they still need to wait 6 seconds before they can throw you, meaning the BSO will not be able to do so sooner compared to the tider, even though they had to manually upgrade their soft grab to a hard grab first.

---

<p>

[Basic Showcase Video](https://github.com/user-attachments/assets/475f5395-f333-40d6-9a99-6015b476a4af)
> _DISCLAIMER: This video will have a timer in red text at the top of the screen, this is added via OBS, and does not exist in-game. Furthermore, I have to manually pause, un-pause, and reset it, so it is not 100% accurate, but it does give you rough idea._

</p>
</details>

---

# Changelog

:cl: BramvanZijp
- add: Initiating a grab on someone now starts a 6 second cooldown before you can throw them.
- tweak: The stamina damage taken upon being thrown has been reduced from 120 to 100.

### DIFF
--- a/Content.Shared/Movement/Pulling/Systems/PullingSystem.cs
+++ b/Content.Shared/Movement/Pulling/Systems/PullingSystem.cs
@@ -320,6 +320,16 @@ public sealed class PullingSystem : EntitySystem
             || component.GrabStage <= GrabStage.Soft)
             return;
 
+        if (_timing.CurTime < component.WhenCanThrow)
+        {
+            args.Cancel();
+            _popup.PopupEntity(Loc.GetString("popup-grab-throw-fail-cooldown",("puller", Identity.Entity(uid, EntityManager)),("pulled", Identity.Entity(args.BlockingEntity, EntityManager))),
+            args.BlockingEntity,
+            uid,
+            PopupType.MediumCaution);
+            return;
+        }
+
         var distanceToCursor = args.Direction.Length();
         var direction = args.Direction.Normalized() * MathF.Min(distanceToCursor, component.ThrowingDistance);
 
@@ -919,6 +929,10 @@ public sealed class PullingSystem : EntitySystem
             _ => throw new ArgumentOutOfRangeException(),
         };
 
+        if (puller.Comp.GrabStage == GrabStage.No)
+        {
+            puller.Comp.WhenCanThrow = _timing.CurTime + puller.Comp.ThrowDelayOngrab;
+        }
         var newStage = puller.Comp.GrabStage + nextStageAddition;
         var ev = new CheckGrabOverridesEvent(newStage); // guh
         RaiseLocalEvent(puller, ev);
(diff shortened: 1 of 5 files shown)


Timfa2112 (approved):

Content.Shared/Movement/Pulling/Systems/PullingSystem.cs
```diff
@@ -919,6 +929,10 @@ public bool TryGrab(Entity<PullableComponent?> pullable, Entity<PullerComponent?
             _ => throw new ArgumentOutOfRangeException(),
         };
 
+        if (puller.Comp.GrabStage == GrabStage.No)
+        {
+            puller.Comp.WhenCanThrow = _timing.CurTime + puller.Comp.ThrowDelayOngrab;
+        }
```
Erisfiregamer1, line 935: If you have CQC you can instantly hard grab which would skip this check and allow for an instant throw anyway. You might want to find a different way to check if they had just initiated the grab.

BramvanZijp:
Gimme a sec to implement the AI suggestions

Erisfiregamer1 (changes requested):
Code looks good to me, but there's an issue with how you're checking if the grab has just been initiated.

Content.Shared/Movement/Pulling/Systems/PullingSystem.cs
BramvanZijp: This is not true, as seen in the video. puller.Comp.GrabStage is the CURRENT grab stage, not newStage, which is made with the code below. This means it checks that before trying to grab them, they were not combat grabbed, and if that is the case, you start the cooldown.

Erisfiregamer1: Oh. Sorry, I don’t know much how the CQC system works, I assumed the hardgrab immediately set it to hard.

Erisfiregamer1 (approved):
Yeah this is fine, my review was wrong

PR #2055, opened 2025-03-22
Grab Changes

BramvanZijp:
# Description

Port github.com/Goob-Station/Goob-Station/pull/1954 from LRP.
Reduce the 6 second throw cooldown to 2 seconds.
The throw cooldown will now reset if upgrading from a soft grab to a hard grab.
CQC Pressure will deal stamina damage AFTER trying to disarm now, while also making sure the popup is always shown.

---

# Changelog

:cl: BramvanZijp
- tweak: Grab Throws now deal less damage as opposed to previously dealing 10 damage to every limb, they also deal less stamina damage.
- tweak: Martial arts will no longer enable instant hard-grabs unless the target is downed, instead allowing you to upgrade grabs twice as fast.
- tweak: Corporate Judo's Judo slam and Armbar now do more stamina damage.
- fix: Sleeping carp now has its proper abilities when the target is downed.
- fix: CQC Kick now also works on people who are standing up, as originally intended.
- tweak: The cooldown before throwing somebody that has been grabbed has been reduced to 2 seconds, but now only starts after hard-grabbing.
- fix: CQC's pressure combo will now try to take the opponents held item BEFORE applying stamina damage instead of after, when they may have already dropped the item due to the stamina damage.


BramvanZijp:
No clue wtf that first commit is doing here but I dont like it.

Content.Shared/_Goobstation/MartialArts/SharedMartialArtsSystem.CQC.cs
```diff
@@ -118,25 +118,29 @@ private void OnCQCSlam(Entity<CanPerformComboComponent> ent, ref CqcSlamPerforme
...
         var mapPos = _transform.GetMapCoordinates(ent).Position;
         var hitPos = _transform.GetMapCoordinates(target).Position;
         var dir = hitPos - mapPos;
         dir *= 1f / dir.Length();
+
+        if (downed)
+        {
+            if (TryComp<StaminaComponent>(target, out var stamina) && stamina.Critical)
+                _status.TryAddStatusEffect<ForcedSleepingComponent>(target, "ForcedSleep", TimeSpan.FromSeconds(10), true);
+            DoDamage(ent, target, proto.DamageType, proto.ExtraDamage, out _, TargetBodyPart.Head);
```
Timfa2112, line 133:
```suggestion
                _status.TryAddStatusEffect<ForcedSleepingComponent>(target, "ForcedSleep", TimeSpan.FromSeconds(10), true);
                
            DoDamage(ent, target, proto.DamageType, proto.ExtraDamage, out _, TargetBodyPart.Head);
```
I like some blank line for readability

Content.Shared/_Goobstation/MartialArts/SharedMartialArtsSystem.CQC.cs
```diff
@@ -157,16 +162,13 @@ private void OnCQCPressure(Entity<CanPerformComboComponent> ent, ref CqcPressure
             || !TryUseMartialArt(ent, proto.MartialArtsForm, out var target, out _))
             return;
 
+        if (_hands.TryGetActiveItem(target, out var activeItem)) // I know this looks horrible, but the disarm should happen BEFORE the stam dmg, and the popup should always show.
+            if(_hands.TryDrop(target, activeItem.Value))
+                if (_hands.TryGetEmptyHand(target, out var emptyHand))
+                    if(_hands.TryPickupAnyHand(ent, activeItem.Value))
+                        _hands.SetActiveHand(ent, emptyHand);
```
Timfa2112, line 169:
```suggestion
        if (_hands.TryGetActiveItem(target, out var activeItem) // I know this looks horrible, but the disarm should happen BEFORE the stam dmg, and the popup should always show.
            && _hands.TryDrop(target, activeItem.Value)
            && _hands.TryGetEmptyHand(target, out var emptyHand)
            && _hands.TryPickupAnyHand(ent, activeItem.Value))
                _hands.SetActiveHand(ent, emptyHand);
```
Double check this, but it's a lit better than if() if() if()

Content.Shared/_Goobstation/MartialArts/SharedMartialArtsSystem.cs
```diff
@@ -176,15 +181,21 @@ private bool TryGrant(GrantMartialArtKnowledgeComponent comp, EntityUid user)
         {
             var canPerformComboComponent = EnsureComp<CanPerformComboComponent>(user);
             var martialArtsKnowledgeComponent = EnsureComp<MartialArtsKnowledgeComponent>(user);
-            LoadPrototype(user, martialArtsKnowledgeComponent, comp.MartialArtsForm);
+            var pullerComponent = EnsureComp<PullerComponent>(user);
+            if (!_proto.TryIndex<MartialArtPrototype>(comp.MartialArtsForm.ToString(), out var martialArtsPrototype))
+                return false;
+            martialArtsKnowledgeComponent.MartialArtsForm = martialArtsPrototype.MartialArtsForm;
```
Timfa2112, line 187:
```suggestion
                return false;
                
            martialArtsKnowledgeComponent.MartialArtsForm = martialArtsPrototype.MartialArtsForm;
```
nitpick

Timfa2112 (changes requested):
A few minor comments
If these are resolved while I'm still asleep, any other maintainer can dismiss my review

Timfa2112:
Also, it'd be real nice to have some media for this so we can see that this all actually works

Content.Shared/_Goobstation/MartialArts/SharedMartialArtsSystem.CQC.cs
BramvanZijp: Fair

Content.Shared/_Goobstation/MartialArts/SharedMartialArtsSystem.CQC.cs
BramvanZijp: Agreed

Content.Shared/_Goobstation/MartialArts/SharedMartialArtsSystem.cs
BramvanZijp: Blame whoever made it for LRP.

BramvanZijp:
> Also, it'd be real nice to have some media for this so we can see that this all actually works

Ill get on that

BramvanZijp:
Applied and tested, it works

PR #2056, opened 2025-03-22
Return Of Bionic Leg

VMSolidus:
# Description

This PR marks the return of "Speed Legs Trait" by adding a new class of trait function for replacing bionic limbs. To be honest this actually needs to be done via Loadouts but whatever. I also made Bionic Arm replace your arm with an actual JOL-Arm. Which means that players can now use surgery to remove the bionics you obtained in character creation.

<p>

image

</p>
</details>

---

# Changelog

:cl:
- add: Added back Bionic Legs. The trait has been reworked so that it makes you start with one of your legs replaced with an actual speed legs cyberware. 
- add: Bionic Arm now gives you an actual bionic arm. Have fun mugging people for their cyberware so you can have a collection of fancy bionic limbs.

### DIFF
--- a/Content.Server/Traits/TraitSystem.Functions.cs
+++ b/Content.Server/Traits/TraitSystem.Functions.cs
@@ -23,6 +23,9 @@ using Content.Shared.NPC.Systems;
 using Content.Shared.Weapons.Melee;
 using Robust.Shared.Audio;
 using Content.Shared.Tag;
+using Content.Shared.Body.Part;
+using Content.Server.Body.Systems;
+using Content.Shared.Body.Components;
 
 namespace Content.Server.Traits;
 
@@ -688,7 +691,7 @@ public sealed partial class TraitAddTag : TraitFunction
 {
     [DataField, AlwaysPushInheritance]
     public List<ProtoId<TagPrototype>> Tags { get; private set; } = new();
-    
+
     public override void OnPlayerSpawn(EntityUid uid,
         IComponentFactory factory,
         IEntityManager entityManager,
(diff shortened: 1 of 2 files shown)


Raikyr0:
Hey uh, TCJ, Both legs need to be Speed Legs for the Speed Bonus. (Or all 8 in the case of an Arachne)

VMSolidus:
> Hey uh, TCJ, Both legs need to be Speed Legs for the Speed Bonus. (Or all 8 in the case of an Arachne)

Fixed.

Content.Server/Traits/TraitSystem.Functions.cs
```diff
@@ -698,3 +701,58 @@ public override void OnPlayerSpawn(EntityUid uid,
...
+    public override void OnPlayerSpawn(EntityUid uid,
+        IComponentFactory factory,
+        IEntityManager entityManager,
+        ISerializationManager serializationManager)
+    {
+        var bodySystem = entityManager.System<BodySystem>();
+        var transformSystem = entityManager.System<SharedTransformSystem>();
+
+        if (!entityManager.TryGetComponent(uid, out BodyComponent? body)
+            || !entityManager.TryGetComponent(uid, out TransformComponent? xform)
```
Timfa2112, line 732:
```suggestion
            || !entityManager.TryGetComponent(uid, out TransformComponent? transform)
```
This doesn't seem like a clear abbreviation--code is _read_ more often than it is written, so abbreviating like this might not be helpful for other contributed seeking to abuse this system later

Content.Server/Traits/TraitSystem.Functions.cs
```diff
@@ -698,3 +701,58 @@ public override void OnPlayerSpawn(EntityUid uid,
...
+            if (partComp.Symmetry != PartSymmetry)
+                continue;
+
+            foreach (var child in bodySystem.GetBodyPartChildren(part.Id, part.Component))
+                entityManager.QueueDeleteEntity(child.Id);
+
+            transformSystem.AttachToGridOrMap(part.Id);
+            entityManager.QueueDeleteEntity(part.Id);
+
+            var newLimb = entityManager.SpawnAtPosition(ProtoId, xform.Coordinates);
```
Timfa2112, line 753:
```suggestion
            var newLimb = entityManager.SpawnAtPosition(ProtoId, transform.Coordinates);
```
This doesn't seem like a clear abbreviation--code is _read_ more often than it is written, so abbreviating like this might not be helpful for other contributed seeking to abuse this system later

Timfa2112 (changes requested):

Content.Server/Traits/TraitSystem.Functions.cs
Erisfiregamer1: isn't xform standard for the transformcomponent?

VMSolidus: Yea, I was about to say this is actually a standard abbreviation for this particular example.

Timfa2112 (approved):
I still don't like the xform abbreviation, but it's not important enough to block merging

PR #2082, opened 2025-03-25
Makes Operating Table Buildable From G Menu Instead of Being a Machine

CometVoid:
# Description

The operating table is currently a machine that requires a frame and a board to craft.
This is a table. Why does it need this? Why does it have the same cost as an autodoc?

It now instead is buildable from the G menu, for the cost of 5 steel and 2 plastic.
Maints surgeons rejoice!

---

<p>

Example Media Embed

</p>
</details>

---

# Changelog

:cl:
- tweak: operating tables can now be assembled from the G menu for the cost of 5 steel and 2 plastic


Timfa2112:
@OldDanceJacket Do you agree with the construction costs for the operating table?

OldDanceJacket (approved):

PR #2084, opened 2025-03-26
Disable Healing Word Revive

GNUtopia:
# Description

As it turns out, Healing Word can actually revive people. If we're giving it roundstart as a trait, that should probably be removed.

---

# Changelog

:cl:
- tweak: Made Healing Word no longer revive


Timfa2112 (approved):
Good PR
Nice and short

PR #2096, opened 2025-03-27
Add Traits to Increase/Decrease Your Carrying Speed & Character Weight.

BramvanZijp:
# Description

Adds 4 new traits.
Bulky (Costs 6 pts) increases your characters density by 55, which increases your weight.
Featherweight (Gives 3 pts) halves your density, halving your weight.

Bodybuilder and Weakling adjust the modifiers for pulling speed, also costing 6 and giving 3.

Also, added a VVModify function for traits, which allows you to get the current value of a VV path, and then change it via multiplication or addition/subtraction, this was needed for Bulky/Lightweight, since Density is under FixturesComponent, which is a robust toolbox thing, and I dont think an engine PR with a densitymodifier field is the way to go.

This was done for a Goob MRP bounty.

---

# Changelog

:cl: BramvanZijp
- add: Added the Bulky and Featherweight traits, which increase/decrease your characters weight.
- add: Added the Bodybuilder and Weakling traits, which impact the movement speed penalty while pulling.


Timfa2112 (approved):
This looks fine to me
@VMSolidus while the TraitVVModify looks alright to me, can you confirm that we want to be able to do this sort of thing in code?

Content.Server/Traits/TraitSystem.Functions.cs
```diff
@@ -297,16 +297,52 @@ public override void OnPlayerSpawn(EntityUid uid,
...
+        foreach (var (path, value) in Changes)
+        {
+            idpath = path.Replace("$ID", uid.ToString());
+            if (!float.TryParse(vvm.ReadPathSerialized(idpath), out var curval))
+                continue;
+            if (Multiply)
+                newval = curval * value;
+            else
+                newval = curval + value;
+            vvm.WritePath(idpath, newval.ToString());
```
sleepyyapril, line 344: please, please, name your variables properly and fully.

Content.Server/Traits/TraitSystem.Functions.cs
```diff
@@ -297,16 +297,52 @@ public override void OnPlayerSpawn(EntityUid uid,
...
-        foreach (var (path, value) in VVEdit)
-            vvm.WritePath(path, value);
+        string idpath;
+        foreach (var (path, value) in Changes)
+        {
+            idpath = path.Replace("$ID", uid.ToString());
+            vvm.WritePath(idpath, value);
+        }
+    }
+}
```
sleepyyapril, line 315: same here.

Resources/Prototypes/Traits/physical.yml
```diff
@@ -1138,6 +1138,93 @@
...
+  points: 3
+  requirements:
+    - !type:CharacterTraitRequirement
+      inverted: true
+      traits:
+        - Bulky
+  functions:
+    - !type:TraitVVModify
+      changes:
+        /entity/$ID/Fixtures/Fixtures[fix1 ]/Density: 0.5
```
sleepyyapril, line 1173: is the space intentional?

sleepyyapril (changes requested):

Content.Server/Traits/TraitSystem.Functions.cs
sleepyyapril: i also don't believe you NEED to specify the "idpath" outside the for loop?

Content.Server/Traits/TraitSystem.Functions.cs
BramvanZijp: Willdo pt2 electric boogaloo

Resources/Prototypes/Traits/physical.yml
BramvanZijp: What space?

VMSolidus: /entity/$ID/Fixtures/Fixtures[fix1 ]/Density: 0.5

There's a space after the 1 in fix1

BramvanZijp: Intentional, thats how the VV path works for... reasons

Content.Server/Traits/TraitSystem.Functions.cs
```diff
@@ -297,16 +297,52 @@ public override void OnPlayerSpawn(EntityUid uid,
...
     {
         var vvm = IoCManager.Resolve<IViewVariablesManager>();
-        foreach (var (path, value) in VVEdit)
-            vvm.WritePath(path, value);
+        string idPath;
+        foreach (var (path, value) in Changes)
+        {
+            idPath = path.Replace("$ID", uid.ToString());
+            vvm.WritePath(idPath, value);
+        }
```
VMSolidus, line 313:
```suggestion
        foreach (var (path, value) in Changes)
        {
            var idPath = (string) path.Replace("$ID", uid.ToString());
            if (!string.TryParse(vvm.ReadPathSerialized(idPath), out var currentValue))
                continue;
            vvm.WritePath(idPath, value);
        }
```
You can try it without the explicit cast as a (string), but this would fix the compiler error you had with it not intuiting that path was factually a string.

Content.Server/Traits/TraitSystem.Functions.cs
```diff
@@ -297,16 +297,52 @@ public override void OnPlayerSpawn(EntityUid uid,
...
+        foreach (var (path, value) in Changes)
+        {
+            idPath = path.Replace("$ID", uid.ToString());
+            if (!float.TryParse(vvm.ReadPathSerialized(idPath), out var currentValue))
+                continue;
+            if (Multiply)
+                newValue = currentValue * value;
+            else
+                newValue = currentValue + value;
+            vvm.WritePath(idPath, newValue.ToString());
```
VMSolidus, line 344:
```suggestion
        foreach (var (path, value) in Changes)
        {
            var idPath = (string) path.Replace("$ID", uid.ToString());
            var newValue = 0f;
            if (!float.TryParse(vvm.ReadPathSerialized(idPath), out var currentValue))
                continue;
            if (Multiply)
                newValue = currentValue * value;
            else
                newValue = currentValue + value;
            vvm.WritePath(idPath, newValue.ToString());
```
This should also be correct.

VMSolidus (changes requested):

Content.Server/Traits/TraitSystem.Functions.cs
BramvanZijp: Done

Content.Server/Traits/TraitSystem.Functions.cs
BramvanZijp: Updated

Timfa2112:
@VMSolidus when you're happy with the PR, so am I, so you can merge it if you approve of it :)

Resources/Prototypes/Traits/physical.yml
```diff
@@ -1138,6 +1138,93 @@
...
+      multiply: true
+    - !type:TraitPushDescription
+      descriptionExtensions:
+        - description: examine-featherweight-message
+          fontSize: 12
+          requireDetailRange: true
+- type: trait
```
VMSolidus, line 1180:
```suggestion
          requireDetailRange: true
          
- type: trait
```

VMSolidus (changes requested):
image

I figured something was not going to work correctly. This system DOES correctly set the density of an entity. HOWEVER it does not update the entity's mass. In order to do so, you would need to call SharedPhysicsSystem.FixtureUpdate on the entity in order to tell the game that it needs to recalculate the entity's physics component. (Modifying the physics component directly is off-limits by the game engine, you can only modify it by calls to the PhysicsSystem). 

For the mass modifying traits, you are going to need to make a dedicated function for this, instead of using direct VVEdit.

VMSolidus (approved):
Alright now it works.

Content.Server/Traits/TraitSystem.Functions.cs
```diff
@@ -556,6 +597,30 @@ public override void OnPlayerSpawn(EntityUid uid,
...
+    {
+        var physicsSystem = entityManager.System<SharedPhysicsSystem>();
+        if (!entityManager.TryGetComponent<FixturesComponent>(uid, out var fixturesComponent))
+            return;
+
+        var fixture = fixturesComponent.Fixtures.First();
+        var newDensity = Multiply ? fixture.Value.Density * DensityModifier : fixture.Value.Density + DensityModifier;
+        physicsSystem.SetDensity(uid, fixture.Key, fixture.Value, newDensity);
+    }
+}
```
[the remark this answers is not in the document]
BramvanZijp: This will not happen, and if we do encounter entities without fixtures, the trait system will be the LEAST of our worries.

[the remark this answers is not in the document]
VMSolidus: Technically the bot is actually right about this. So I'm accepting his commitable suggestion. It's not a bad idea to accept safety checks like this.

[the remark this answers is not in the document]
VMSolidus: I will however say to @coderabbitai that our repo coding standards are that multiple "guard clause" checks like this should be combined into a single guard clause using the || operator. And that we prefer said operator to be to the left of the check on the same line, in the same indentation as the return.

PR #2098, opened 2025-03-27
You Can Use a Screwdriver to Change Energy Sword Colors

Tirochora:
# Description

Energy swords of all sorts can now be screwed to change the color in a specific pattern. They are still randomized on roundstart, but can now be manually adjusted with a screwdriver (or anything that can be used as one). This is dependent on the colors available to the e-sword, too, so the HoS' antique will still only ever be cyan.

---

<p>

esworddescription

</p>
</details>

---

# Changelog

:cl:
- add: You can now use a screwdriver on an energy sword to change what the color of the blade is. This applies to all variants that have color variation.


Tirochora:
I could make this better.

PR #2102, opened 2025-03-28
Misc Changes

EctoplasmIsGood:
# Description

Decreases weights on various animal spawn events so that interesting stuff actually happens again, removes pAI from loadout so having one is actually cool and there aren't 5 zillion ghost roles, makes greenshift vote only so it can be used for events but doesn't make rounds boring by getting a greenshift with no admins on, makes hellshift more violent (hopefully), and adds a roundstart event for changelings so they can actually spawn.

---
# Changelog

:cl:
- add: Changelings can now spawn as a game mode. don't know why they couldn't before.
- remove: Greenshift can no longer randomly happen; it has to be voted for.
- remove: No more loadout pAIs
- tweak: Animal spawn events are rarer
- tweak: Random Sentience event is rarer
- tweak: Hellshift should be more hellish

### DIFF
--- a/Resources/Prototypes/GameRules/events.yml
+++ b/Resources/Prototypes/GameRules/events.yml
@@ -244,7 +244,7 @@
   components:
   - type: StationEvent
     earliestStart: 30
-    weight: 7.5
+    weight: 5
     minimumPlayers: 10 #Enough to hopefully have at least one engineering guy
     startAnnouncement: true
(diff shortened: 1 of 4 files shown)


Mike32oz:
It's not recommended to remove the green shift

EctoplasmIsGood:
I have an idea

VMSolidus (approved):

PR #2110, opened 2025-03-29
Tweak IPC's Damage Resistances to Reduce Their Extreme Weaknesses

BramvanZijp:
# Description

Title:
Heat damage multiplier: 2 > 1
Shock damage Multiplier: 2.5 > 2

While I get that in terms of UTILITY, IPCs are in a powerfull state, they suck at combat. EMPs, double heat, 2.5x shock, and now a dedicated ion damage type is being added to further nerf IPCs in combat, so this is going to become neccesary.

The new modifier set is still a net negative, but nowhere near as severe as it was before.

---

# Changelog

:cl: BramvanZijp
- tweak: Tweaked IPC's damage resistances to no longer take double heat damage, and slightly reduce their shock weakness.

### DIFF
--- a/Resources/Prototypes/Entities/Mobs/Player/silicon_base.yml
+++ b/Resources/Prototypes/Entities/Mobs/Player/silicon_base.yml
@@ -338,5 +338,9 @@
   coefficients:
     Poison: 0
     Cold: 0.2
-    Heat: 2
-    Shock: 2.5
+    Heat: 1.25
+    Shock: 1.5
+    Ion: 1 # They are already the only player species to take this damage at all, so no need to compound that weakness
+    Brute: 1
+    Slash: 0.9
+    Piercing: 0.9
--- a/Resources/Prototypes/Entities/Mobs/Player/silicon_base.yml
+++ b/Resources/Prototypes/Entities/Mobs/Player/silicon_base.yml
@@ -338,9 +338,9 @@
   coefficients:
     Poison: 0
     Cold: 0.2
-    Heat: 1.25
-    Shock: 1.5
+    Heat: 1
+    Shock: 2
     Ion: 1 # They are already the only player species to take this damage at all, so no need to compound that weakness
     Blunt: 1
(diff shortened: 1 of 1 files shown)


OldDanceJacket:
This is an over correction and seems to be a bit knee-jerky, however I agree with the general idea you have:
 
They do not need a resistance to Slash and Pierce, they already have a resist to these damage types in the fact that they don't bleed(which is the primary strength of slash). 
Revert them back to normal is my thoughts.

Normalizing the heat to be far less extreme as 'balancing' factors is a good choice, outright just remove the heat weakness, they have to use special healing for it and we have a dedicated damage type in Ion MEANT for them.

Change shock from 1.5 -> 2.0 , it's a rather unique damage type and is honestly a bit funny. I think later down the line taking shock damage should/could probably do something unique to IPC like making them say weird things or whatever(like they're on the fritz), if this is too much we can lower it further to 1.75 or 1.5 but lets just lower it a little to start, find a sweet spot.

BramvanZijp:
> This is an over correction and seems to be a bit knee-jerky, however I agree with the general idea you have:
> 
> They do not need a resistance to Slash and Pierce, they already have a resist to these damage types in the fact that they don't bleed(which is the primary strength of slash). Revert them back to normal is my thoughts.
> 
> Normalizing the heat to be far less extreme as 'balancing' factors is a good choice, outright just remove the heat weakness, they have to use special healing for it and we have a dedicated damage type in Ion MEANT for them.
> 
> Change shock from 1.5 -> 2.0 , it's a rather unique damage type and is honestly a bit funny. I think later down the line taking shock damage should/could probably do something unique to IPC like making them say weird things or whatever(like they're on the fritz), if this is too much we can lower it further to 1.75 or 1.5 but lets just lower it a little to start, find a sweet spot.

So remove the pierce/slash resist and make shock 2, thats fine

OldDanceJacket (approved):

PR #2112, opened 2025-03-29
Some Of These Really Dont Fit

VMSolidus:
Some of the Lifepaths were wildly inappropriate for a character that can be employed on the station.


bruhmogus:
i really don't see why cultist is so bad? its pretty implied that people with cultist lifepath escaped or whatever
is it just because blood cultist exists?

OldDanceJacket (approved):

PR #2116, opened 2025-03-29
Let Foxes and Dogs Track Scents

musicman928:
# Description

I thought it was weird that foxes and dogs can't track scents like Vulpines, so I fixed it. That's all there is really. I'm just surprised this wasn't done when Vulpines initially got the ability, it just makes sense.

This makes it so every dog and fox have the ability to track scents when controlled by a player. This includes Laika, Ians, Renault, Siobhan, Waltuh, and Cerberus (since it extends MobCorgi), as well as all foxes and dogs purchased from cargo.

# Changelog

:cl: PsychoLogik (musicman928)
- tweak: Dogs and Foxes can now track scents (Cerberus included)

### DIFF
--- a/Resources/Prototypes/DeltaV/Entities/Mobs/NPCs/animals.yml
+++ b/Resources/Prototypes/DeltaV/Entities/Mobs/NPCs/animals.yml
@@ -67,7 +67,6 @@
     tags:
     - VimPilot
   - type: ScentTracker
-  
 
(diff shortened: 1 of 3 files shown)


Timfa2112 (approved):
Approved assuming @OldDanceJacket has no objections
(I bother him for balance stuff I'm not sure about)

PR #2120, opened 2025-03-30
Push More Blackburn Content so I Don't Forget About It and Accidental…

Plyushsune:
…ly delete my work

# Description
image

image
image

:cl:
- add: Added fun :D
- tweak: Tweaked fun
- fix: Fixed fun!
- remove: Removed fun :(


Plyushsune:
something broke

EctoplasmIsGood:
gunsprite looks weird. bayonet is tiny, colors are super drab, barely even looks like a gun, and doesn't fit in with the rest of the games sprites at all, at least in my opinion. The clothing looks fine, but that gun needs to be completely redone because frankly it just looks bad.

Timfa2112 (changes requested):
The camouflage on that rifle is making it seem a bit flat. You might want to consider a slightly darker blueish colour? I'm not sure

Also your changelog is still using the template, do fix that

Plyushsune:
> The camouflage on that rifle is making it seem a bit flat. You might want to consider a slightly darker blueish colour? I'm not sure
> 
> Also your changelog is still using the template, do fix that

I don't know how to use a changelog and I'm just gonna redo the rifle because everyone doesn't like it

Plyushsune:
i also killed the sk-40

PR #2121, opened 2025-03-30
sunglasses are more expensive

Tirochora:
# Description

This PR changes the sunglasses to be more expensive in loadouts, as I find the fact that they are as cheap as they are to be very unbalanced. As per the suggestion of Parky, I have also renamed the default such that it specifies ones that are rated for flash protection.

---

<p>

sunglasses

</p>
</details>

---

# Changelog

:cl:
- tweak: sunglasses with flash protection are more expensive in the loadouts
- tweak: sunglasses are either "default" or "shielded" rather than "cheap" or "standard." the description has also been changed to reflect this


Timfa2112 (approved):
Seems fine to me, but I'm requesting a second opinion from @VMSolidus to be sure

PR #2142, opened 2025-04-02
Tweak Language Stuff

3PonPon3:
# Description

This PR removes the limit to how many basic language traits you can take, reduces the cost of sign language from -2 to -1 and ensures that you cannot take more than 1 accent from the accent traits. (Previosly, when new accents got added, the bunch with german, french, spanish, italian and russian, they weren't added to the character item group that makes it so you can only take 1, so I added them)

---

# Changelog

:cl:
- tweak: Reduced sign language cost from -2 to -1
- tweak: Removed the limit on how many language traits you can take
- fix: Made sure that you cannot take more than 1 accent trait


Timfa2112 (approved):
This change was already agreed upon by me and ODJ in a different PR, so it seems fine to me

PR #2172, opened 2025-04-05
Comment Our Item Group Req and the Group

3PonPon3:
# Description

Yeah so i fumked up in my previous language related PR and this PR actually makes it so you can take UNLIMITED languages and finally create John Language. It does this by commenting out the basic languages item group, so they no longer have a limit due to not being in the item group with max items set. Also this is now made on the correct branch and no longer has commits from a different PR......oops.....

---

# Changelog

:cl:

- tweak: Commented out the basic languages item group so now you can finally take UNLIMITED language traits and become John Language himself. For real this time.


VMSolidus (changes requested):
Not necessary here: See #2170

3PonPon3:
Shit, he's right. ATTEMPT THREE HERE WE GO

PR #2212, opened 2025-04-10
(Bandaid Fix for Languages/Accents) I'M LANGUAGING OUT MAN!!!

3PonPon3:
# Description

This PR comments out the CharacterItemGroup for basic languages so you can take ALL OF THEM (side effects may include using all your trait slots) and the lines responsible for breaking both languages and accents (the ones that specify how many slots it takes up in the item group and how many trait slots it uses)
This is so we can have them functioning until that code gets fixed

---

# Changelog

:cl:
- fix: Languages and accents now work. JOHN LANGUAGE IS FINALLY REAL!

### DIFF
--- a/Resources/Prototypes/CharacterItemGroups/Generic/languageGroups.yml
+++ b/Resources/Prototypes/CharacterItemGroups/Generic/languageGroups.yml
@@ -1,33 +1,33 @@
-- type: characterItemGroup
-  id: TraitsLanguagesBasic
-  maxItems: 1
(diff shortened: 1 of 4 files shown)


Timfa2112 (approved):

Timfa2112 (changes requested):
Please add media in the PR

Timfa2112 (approved):
Found media on Discord. Put it in the PR next time!

PR #2219, opened 2025-04-11
Revert "Initial Corporate Requirements"

EctoplasmIsGood:
After some discussion in EE, community outrage from MRP, and overall just negative reactions all around, I'm gonna be opening a PR to revert this. 

Locking every role behind specific corporations makes it extremely tedious to switch jobs, and is borderline nonsensical, as it implies nanotrasen has NO doctors of their own, even though they're a megacorp that basically controls a galactic empire. 

All jobs should be selectable by Nanotrasen employees, and this PR goes against that.


Timfa2112 (changes requested):
Your change is also deleting the newly added PMCG—try to filter that out so you're only removing the requirements

VMSolidus (approved):
More merge conflic bait let's go lmao.

EctoplasmIsGood:
> Your change is also deleting the newly added PMCG—try to filter that out so you're only removing the requirements

EctoplasmIsGood:
https://github.com/Simple-Station/Einstein-Engines/pull/2220

PR #2220, opened 2025-04-11
Add Nanotrasen as an Employer for Every Job

EctoplasmIsGood:
# Description

Adds Nanotrasen as an employer for every job, because you are on an NT station. Also makes it so unemployed people can take clown, mime, musician, and janitor.

# Changelog

:cl:
- add: Nanotrasen employees can take every job again


VMSolidus (approved):

VMSolidus:
You forgot the Mantis by the way. :)

PR #2225, opened 2025-04-12
Remove Pulse Pistol From Loadouts

EctoplasmIsGood:
# Description

Another PR to remove the pulse pistol from loadouts, as the previous one was abruptly closed because some maps used to have it mapped. This gun is still insanely overturned, and just because it was mapped for some reason in the past doesn't mean it should continue to be accessible. This is a DEATHSQUAD weapon and is balanced as such.

I should also make it clear this PR was made by request of a Goob MRP admin, and I 100% agree with the request because this weapon is comically strong. 

# Changelog

:cl:
- remove: Captain and HoS can no longer take the pulse pistol from loadouts


EctoplasmIsGood:
The fact this gun was mapped for HoS doesn't unfuck its stats

DEATHB4DEFEAT:
don't do that

PR #2238, opened 2025-04-13
Surgery Experience Traits + CPR Trait Tweak

3PonPon3:
# Description

This PR adds 2 new surgery related traits to the game (surgery training which boosts your surgery speed to 1.6, only available for non-medical jobs, as well as experienced surgeon, which boosts your surgery speed to 2.5, available only to medical jobs + corpsman, without intern. They both cost 3 points.) while also tweaking and fixing the CPR training one. (reduces cost from 4 to 3 and adds BSO to the blacklist as BSO already has CPR training innately) 

---

<p>

surgon

surgon2

base surgery speed

Surgery Training speed

Experienced Surgeon speed

</p>

</details>

---

# Changelog

:cl:
- add: Traits that increase your surgery speed, Surgery Training (1.6, between intern and standard doc) for non-medical, and Experienced Surgeon (2.5, same as CMO) for medical w/o intern and with corpsman
- tweak: CPR Training cost 4 -> 3
- fix: BSO can no longer take the CPR trait as they already have it innately

### DIFF
--- a/Resources/Locale/en-US/traits/traits.ftl
+++ b/Resources/Locale/en-US/traits/traits.ftl
@@ -617,3 +617,15 @@ trait-description-RestrictedGear =
     Either through personal ownership or theft, you have access to equipment that isn't particularly standard issue.
     Note that starting with an item [color=red]doesn't certify its legality[/color]. Conceal it or justify it.
     (You equip other jobs' items in the loadouts menu)
+
+trait-name-SurgeryTraining = Surgery Training
+trait-description-SurgeryTraining =
+    At some point in your life you acquired the knowledge and experience necessary for performing surgery effectively. 
+    This trait boosts your surgery speed to 1.6, and is intended for non-medical characters, as medical jobs already have faster surgery. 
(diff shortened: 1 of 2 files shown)


Timfa2112 (approved):
Good example of clear media in a PR, appreciated!
Pinging @OldDanceJacket for a second set of eyes on balance--the technical side checks out.

3PonPon3:
I'll say that in terms of balance I asked others what they think too, and keep in mind surgery is just a small part of what you do as medical, so I think 3 points is fair for both traits. 
As non-medical the use varies, it can be for maints surgery, it can be for salvage, it has a few uses and can be dealbreaking for some, but overall, not that impactful of a change. 
The one for medical I still consider fine as not everyone is medical is going to take the trait, and it boosts around the same as the trait for non-medical. (Non-medical: 1->1.6, medical 1.75/2->2.5) 
So while it will certainly be impactful, it isn't going to be the meta or OP, as surgery is only a part of medical and healing. Also, you could do maints surgery better for funny.
You can say if you think differently, open to hearing your thoughts if it should be balanced differently.

OldDanceJacket (approved):

PR #2246, opened 2025-04-15
Make Tests Shut Up About Audio Entities

VMSolidus:
# Description

If wizden are okay with doing this then screw it I'll do it to DeleteAllThenGhost too.

https://github.com/space-wizards/space-station-14/pull/34021

# Changelog

No CL this isn't playerfacing but great for my sanity.

### DIFF
--- a/Content.IntegrationTests/Tests/EntityTest.cs
+++ b/Content.IntegrationTests/Tests/EntityTest.cs
@@ -2,6 +2,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Numerics;
 using Robust.Shared;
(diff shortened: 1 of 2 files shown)


ilmenwe:
I had a PR to that one as well when it comes to DeleteAllWhenGhost in debug. 
I increased the duration to wait, but I am unsure if this was the right solution. 
It made the test pass if it was given 150 ticks instead of 5. 

Maybe the duration for things to proparly despawn have increased with time with more logging and more debug validation being processed?

I think i tested it in release and it passed there without a problem.

Timfa2112 (approved):

PR #2247, opened 2025-04-15
Clean Up Tableslam + Strangling Bugfix

Tirochora:
# Description

Previously, the "TryStopPull" function didn't make you drop a virtualitem if you were strangling somebody, meaning that you could lose use of a hand until disarmed because it would still consider you to be holding the person. I fixed that, as well as additionally cleaned up the table slam system, which should make it run a bit smoother and be functionally about the same. The "tableable" and "posttabled" components were removed, because they shouldn't have existed in the first place. The only notable non-bugfix change that's player facing is that shoving people who are knocked down on any climbable entity stuns them (as opposed to just tables), but it's a rather minor change and I intend on reworking it pretty heavily once my [other PR](https://github.com/Simple-Station/Einstein-Engines/pull/2199) is reviewed. On the backend, I added an optional variable to "TryClimb" that gives you an option to skip the do-after. If it's possible to change these actions to be predicted, I'd be interested in learning how.

---

# TODO

- [ ] Task
- [x] Completed Task

---

<p>

Note that glass tables do much more damage when slammed into, and even more if they would shatter when attempting to climb it.

https://github.com/user-attachments/assets/bc0a12e3-0b67-4d61-aa4e-785e3210c3bb

</p>
</details>

---

# Changelog

:cl:
- tweak: Table slamming should be more consistent.
- fix: You should properly be able to use both hands after letting somebody go from a stranglehold.


Tirochora:
> **Actionable comments posted: 0**
> 🧹 Nitpick comments (4)
> 
> > Content.Shared/_Goobstation/TableSlam/TableSlamSystem.cs (4)> `50-51`: **Eliminate commented-out, deprecated code.**
> > > Your loyal soldiers are leaving behind a few unburied corpses of code. Since these old event subscriptions and methods have been fully replaced elsewhere, it’s best to remove them outright for a cleaner codebase. If there's no plan to reintroduce them, consider fully deleting them to avoid confusion.
> > > Apply this diff to remove unnecessary comments:
> > > ```diff
> > > - //SubscribeLocalEvent<PullableComponent, StartCollideEvent>(OnStartCollide);
> > > - //SubscribeLocalEvent<KnockedDownComponent, DisarmAttemptEvent>(OnDisarmAttemptEvent);
> > > 
> > > - // Deprecated, this feature has been moved to SharedStunSystem.cs
> > > - /* private void OnDisarmAttemptEvent(Entity<KnockedDownComponent> ent, ref DisarmAttemptEvent args)
> > > - {
> > > -     ...
> > > - }*/
> > > 
> > > - /*public override void Update(float frameTime)
> > > - {
> > > -     ...
> > > - }*/
> > > 
> > > - /* private void OnStartCollide(Entity<PullableComponent> ent, ref StartCollideEvent args)
> > > - {
> > > -     ...
> > > - }*/
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Also applies to: 54-61, 63-72, 128-158
> > > 
> > > `90-90`: **Confirm rounding behavior for probability.**
> > > Rounding to the nearest quarter is a strategic call. Ensure that you truly intend midpoint rounding to even (for .NET). If you want to bias up or down, adjust accordingly. Otherwise, this is acceptable as is.
> > > 
> > > `104-104`: **Watch for possible stacking of paralyze and knockdown.**
> > > Glass tables inflict both a paralyze (line 108) and a knockdown (line 101). If that’s intentional—aiming to thoroughly punish the victim with multiple overlapping statuses—then proceed. Otherwise, consider unifying them or employing a single effect for consistency.
> > > Also applies to: 106-108
> > > 
> > > `112-125`: **Duplicate calls to damage the same entity.**
> > > The code calls `_damageableSystem.TryChangeDamage(ent, ...)` twice in quick succession with nearly identical specs. You do specify a torso target part in one call but not the other, which may be intentional. If the duplicated damage is fully intended, keep it. Otherwise, consider merging the logic for readability and maintainability.
> 
> 📜 Review details

One of the TryChangeDamage calls is directed to the entity on top of the table, while the other is directed towards the table itself to cause it to shatter. However, I did invert which one should have the TargetBodyPart parameter, as glass tables don't have a torso last I had checked.

VMSolidus (approved):

PR #2272, opened 2025-04-18
Various Misc Psi Power Fixes

VMSolidus:
# Description

The bug with the traits menu related to psionics is a significantly more complicated issue, so I'm fixing that separately, it notably involves touching a different system entirely than anything in this PR. The biggest ones this fixes are Psi-Invis and Anoigo. Others are a variety of other powers that were supposed to be checking "Is the target insulated?" but weren't. To make it easier to check, I added an overload for the OnAttemptPowerUse function that accepts a Target EntityUid, which extends the insulation check to a second entity, which all of the pre-existing targetted powers now use. Which simplifies the need to constantly write the checks over and over again. 

The biggest fixes this patch from to Psi Invisibility, and to an underlying secondary problem that I discovered as a result of more closely examining the cause of the Invis bugs. So now the client gets networked their casting stats. Both Psi-invis and Anoigo have had their audio fixed (it was a typo in both cases). Anoigo now also checks if the target has insulation.

This last point is relevant because certain "High Security" doors have had their components psifoil insulated to protect from people with a rare and obnoxious power from forcing their way in. 

<p>

https://github.com/user-attachments/assets/ba53df07-5ed9-4b08-b787-a6f6b347c8a3

</p>
</details>

# Changelog

:cl:
- add: NT has had its station's "High Security" doors such as Command, Security, and Vault doors shielded with psifoil around their components. They can no longer be forced open by the Anoigo power. 
- tweak: Anoigo, Dispel, Mindswap, Healing Word, Breath Of Life, Pyrokinesis, and Noospheric Zap now all don't work if the target has psionic insulation.
- fix: Psionic Invisibility now actually works as advertised. 
- fix: Psionic Invisibility and Anoigo now actually produce a sound, as well as properly enforce their cooldowns.
- tweak: Anoigo reclassified as a "Dangerous" power, it now also costs 10 points for Elementalists.
- remove: Prisoners can no longer be Psionic.
- fix: Mantis can no longer summon his black blade if his hands are full.


Timfa2112 (approved):
We'd need to test on a Linux server to know for sure, but as far as I can tell the changes work and look good. Well done!

PR #2289, opened 2025-04-19
Fix Friction without Input

DEATHB4DEFEAT:
Problem I had on Parkstation, ported the fix here.


VMSolidus (approved):
image

Technically the documentation comment here is incorrect now lul.

Erisfiregamer1 (approved):
Yeah.

PR #2300, opened 2025-04-22
Integrated Jetpacks For Syndicate Modsuits

Orion-416:
# Description

Adds integrated jetpacks for both of the syndicate modsuits. For now both don't consume any gas, due to a weird bug where anything that has storage in it, and also is a gas tank, can't be inserted into a gas canister.

---

# TODO

- [x] Add jetpacks to both syndie modsuits

---

<p>

image

image

</p>
</details>

---

# Changelog

:cl:
- tweak: Both Syndicate modsuits now have integrated jetpacks


EctoplasmIsGood:
this should be a draft til you fixthe gas usage

Orion-416:
The elite modsuit was meant to be something close to an ion jetpack, but we don't really have support for that until modsuit modules, and I don't know how to fix the issue with gas (i'm pretty sure it's something to do with it having storage, which for some reason makes it unable to be inserted into a gas cylinder)

VMSolidus (commented):
These should actually be on the chestpiece.

PR #2304, opened 2025-04-23
Disable Lavaland Worldgen In Debug

VMSolidus:
# Description

This ONLY prevents the tests from inappropriately generating lavaland constantly, which lags them the hell out and kills them constantly. Consider it the same reason why Worldgen is also disabled on Debug. This notably also doesn't prevent the Lavaland test from running, nor does it prevent the tests from verifying that the lavaland related grids don't have issues. It just prevents the stupid lavaland map itself from generating in completely unrelated tests. If you don't believe me, go watch the test timer and enjoy it not taking 45 minutes anymore.

# Changelog

No CL this isn't player facing.


MLGTASTICa:
Does this disable it in Debug config for Content.Server , or is this just for Tests really.

PR #2310, opened 2025-04-25
Platelet Factories Buff 2

Raikyr0:
# Description

We're buffing platelets yet again. Why? Auto Repair bots still wipes the floor with Platelets.
So, I did some minor chaotic changes to Platelets.

Instead of healing overall of one damage type, it heals every damage type under that, so on paper if you were to take a wide spectrum of damage it would heal drastically faster. Of course you'd still eat shit if you were to be mag dumped.

Perhaps the biggest buff in there is the separation of Bloodloss and Asphyxiation which should drastically slow your death in crit should you also be suffering bloodloss.

Also, fucking around and seeing if I can use some other coding to fill blood back in. Numbers might be a tad high but it's gotta compete with the IPC one.

# TODO

- [x] Figure out how to make it so Platelets fills blood back in

# Changelog

:cl:
- add: Blood Regeneration to Platelets
- tweak: Oh god why the fuck did you do this


Raikyr0:
I have a vague guess where the Linter Issue is.

Raikyr0:
Yep, the Linter Issue was entirely because I had groups: instead of types:

Raikyr0:
> This just isn’t something we need at all. Nano repair factory has high numbers because IPCs have all their healing concentrated onto fewer types, and have a smaller health pool. IPCs are meant to be easier to fix.

choosing to ignore this simply because I want to inch this closer to Nano Repair Factory power level, the average joe really does just go, wow Nano Repair Factory is insane, and Platelet Factories is no where close, more so now they're lowering the amount of points you can play with.

Raikyr0:
I can confirm it's actually working and you will regenerate 2.5% blood assuming you are not hungry or thirsty. Numbers are still adjustable.

EctoplasmIsGood:
> > This just isn’t something we need at all. Nano repair factory has high numbers because IPCs have all their healing concentrated onto fewer types, and have a smaller health pool. IPCs are meant to be easier to fix.
> 
> choosing to ignore this simply because I want to inch this closer to Nano Repair Factory power level, the average joe really does just go, wow Nano Repair Factory is insane, and Platelet Factories is no where close, more so now they're lowering the amount of points you can play with.

We expressly don’t want platelets to be as strong as nanorepair. One of IPCs main gimmicks is their ease of healing.

Raikyr0:
> > > This just isn’t something we need at all. Nano repair factory has high numbers because IPCs have all their healing concentrated onto fewer types, and have a smaller health pool. IPCs are meant to be easier to fix.
> > 
> > 
> > choosing to ignore this simply because I want to inch this closer to Nano Repair Factory power level, the average joe really does just go, wow Nano Repair Factory is insane, and Platelet Factories is no where close, more so now they're lowering the amount of points you can play with.
> 
> We expressly don’t want platelets to be as strong as nanorepair. One of IPCs main gimmicks is their ease of healing.

Either way I asked Timfa and there's some major rebalance stuff going on, so we'll find out eventually.

Diggy0:
Test your PRs or I will send a Deathsquad to your house.

VMSolidus (changes requested):
God no, please no. This is terrible. We already have problems with traits being comically excessive, and this is just quadrupling the speed at which platelet factories heals you, when it was already made 5x faster than it was in its original incarnation to begin with.

Raikyr0:
> Test your PRs before sending em or I will send a Deathsquad to your house.

Actually Tested btw and the blood regen thing actually works assuming you aren't hungry or thristy

> God no, please no. This is terrible. We already have problems with traits being comically excessive, and this is just quadrupling the speed at which platelet factories heals you, when it was already made 5x faster than it was in its original incarnation to begin with.

lets be fair the original iteration of platelets are so dogshit bad it was a 10 cost trap, but yeah i should tune it down slightly, the testing I did proved that if you took like 4 damage types you could heal out of crit. (while thematically cool, a little strong)

Raikyr0:
Numbers have been decreased across the board which incidentally makes it worse against pure single damages.

Raikyr0:
Yeah ok I just saw the Nano Healbot nerf I suppose I should tune this down slightly more.

GNUtopia:
> Numbers have been decreased across the board which incidentally makes it worse against pure single damages.

that’s not how damage groups work
for example, the current platelets ALWAYS heal 0.117 slash per second irrespective of whether you only have one brute type or all three
this is strictly stronger than before

also if you’re using the blood deficiency component for this you should make the traits mutually exclusive so they don’t conflict

Raikyr0:
> > Numbers have been decreased across the board which incidentally makes it worse against pure single damages.
> 
> that’s not how damage groups work for example, the current platelets ALWAYS heal 0.117 slash per second irrespective of whether you only have one brute type or all three this is strictly stronger than before
> 
> also if you’re using the blood deficiency component for this you should make the traits mutually exclusive so they don’t conflict

sorry my brain was cooked from migraine at the time, but yeah it should heal stronger now, with how many points one receives from negative traits i don't think it'll be too much of an issue as a big cap stone trait

Also amazingly blood deficiency didn't conflict, it simply made it so you'd regenerate (a minuscule) less blood, I've made them exclusive so it's not a free pick if you had the blood deficiency trait tho.

PR #2323, opened 2025-04-26
Trait Repricing, 2! 

OldDanceJacket:
# Price drops and value cuts

Based off of player feedback and overall issues with the design of points following previous PR's, prices for a big chunk of traits have been adjusted. 
This has very intentional direction regarding traits.

Firstly, a number of negative traits have been reduced in the amount of points they return to the user. This is simply a reduction of points however, as no negative trait was changed to give 0, with a few of the more 'awful' negative traits given a few extra points for those that opt in. 

Next up is the various price reductions across the board to a number of the under-preforming positive traits. A number of these were rather interesting but were hampered by silly prices. 

Aside from this, a few traits fall into a slight balance adjustment, such as vampirism becoming more expensive, a slight nerf to the regen of Nanobots(nothing else yet), and the adjustment to the Oni's trait selection no longer costing them any trait points. 

Please enjoy!

---

<p>

rotating

</p>
</details>

---

# Changelog

:cl: ODJ
- tweak: Adjusted a number of trait prices; Oni now have a free species trait choice, Harpies singing is far cheaper, being small is now free, etc.


3PonPon3:
I see you turned down the experienced surgeon bonus. Did you reflect it in the description? As it has the exact value there.

Timfa2112 (approved):

PR #2339, opened 2025-04-27
More Random Code Optimizations

VMSolidus:
# Description

These are all essentially just random systems that scaled directly with player count (or were on my list of the top 50 anyways). So just a couple systems that had very inefficient enumerators being swapped out with significantly more efficient ones, plus a few swaps from O(n) to O(m) m << n. A big one was DrainSystem, which was querrying all possible entities, rather than doing so from the "set of all static objects", which is significantly smaller. Puddles are always static objects, so it doesn't make sense to have the drains check for anything other than static. 

We can also use DirtyField to save on performance costs of Dirty(uid, component) in cases where the Dirty is only networking a single component field. 

# Changelog

no CL this isn't player facing.


DEATHB4DEFEAT (approved):

VMSolidus:
Aww man, I was hoping Psprite would come see that I exclusively used the ActiveComponent pattern here that he told me about.

PR #2343, opened 2025-04-28
Remove Chitinids

VMSolidus:
# Description

These awful shitty things were 5% of all frametime on any server that had then enabled. They are singularly responsible for also making it so that RadiationSystem even appears at all in the top 20 systems (It was previously not even in the top 100 in my dev environment). 

# Changelog

:cl:
- remove: Removed Chitinids.


OldDanceJacket (approved):

Timfa2112 (changes requested):
Would it not be far better to disable them than to delete them outright? CCVAR or otherwise. 
This way downstreams get no say in whether or not they want to keep them.

Greenwall-Exe:
could we just remove the rad nonsense and replace it with different perk? like maybe a second backpack slot to show how ants can carry heavy stuff or something? I was just recently beginning to see more of them, hate to see them get deleted wholesale

ZetaNull:
I think the idea was floated of reworking them into being moth markings seeing as the sprite is basically just an edit from what it looks like

The way theyre actually coded is horrifying and was apparently managing to make them take almost as much processing power as the _atmos system_ on goob mrp though, on top of emitting radiation and having a built in light among other things if im remembering that right

sleepyyapril:
just optimize it lol

Theracter:
I hate this a lot and would rather them be rewritten

mi8hip:
I do like the performance boost this would give, but completely removing the species seems excessive. Maybe just disable them until someone wants to rework the abilities?

Erisfiregamer1 (changes requested):
mald pr

Optimizing the species instead of removing them would be a better move here. I know you want to make this codebase run fast, but this ain't it chief.
Maybe figure out why RadiationSystem is so slow with them?

ZetaNull:
I've just got chitinids to work as moth markings, at least in terms of their visuals. If we opt to go the route of merging that with moths, might be good to give them a trait too for _some_ of what they did mechanically? At least the stuff that wasn't horrifyingly unoptimized shitcode? Unsure.

Might be better to disable than outright remove them but they seem kind of. Scuffed. With a lot of the ways they're implemented. So moth markings until someone feels like optimizing them properly might be a decent play

VMSolidus:
Gunna be replacing this PR with a lighter one that just makes them default to non-roundstart with a big ass warning label about their status as server performance killers.

VMSolidus:
> I've just got chitinids to work as moth markings, at least in terms of their visuals. If we opt to go the route of merging that with moths, might be good to give them a trait too for _some_ of what they did mechanically? At least the stuff that wasn't horrifyingly unoptimized shitcode? Unsure.
> 
> Might be better to disable than outright remove them but they seem kind of. Scuffed. With a lot of the ways they're implemented. So moth markings until someone feels like optimizing them properly might be a decent play

Actually they had two C# systems, and neither of them are salvagable in their current state. The ChitinidComponent needs to be completely disabled for "UNGODLY SHITCODE REASONS"

While InjectionBlockerComponent needs to be refactored into something NOT massively hardcoded to only be useful for Chitinids.

ZetaNull:
Moreso was meaning things like some of their resistances or density in terms of what they did mechanically, I think ODJ had a trait planned out for that

PR #2351, opened 2025-04-28
Equipment Tweaks

EctoplasmIsGood:
# Description

Two minor changes. First off, civilian disabler is now a command loadout item, not an item for every passenger on the planet, and as such, it's been moved to the sec fabricator instead. Second off, there is no longer a judo belt in EVERY LOCKER., The sec tech spawns with two, so they get to fight over them.

image
---

# Changelog

:cl:
- tweak: Civillian disablers are now a command loadout item, and shouldn't be used by random people. 
- tweak: Judo belts don't spawn in sec lockers anymore, now the sec vend starts with two in its stock.


EctoplasmIsGood:
why is there so many random commits???

OldDanceJacket:
Your PR lacks media, please provide media of something working here.

Similarly is this what we want to do going forward? Is this an agreed upon direction?

OldDanceJacket (changes requested):
Provide media.

EctoplasmIsGood:
> Provide media.

brother its a removal media of what

PR #2366, opened 2025-05-01
Equipment Tweaks

EctoplasmIsGood:
# Description

It removes civilian disablers from everybody's loadouts, instead of giving them commands to combat valid hunting and unfun antag gameplay. It also reduces the number of judo belts sec gets, removing them from sec lockers and putting two in the sec tech instead.

<p>

image

</p>
</details>

---

# Changelog

:cl:
- tweak: Civillian disablers are now a command loadout item, and shouldn't be used by random people.
- tweak: Judo belts don't spawn in sec lockers anymore, now the sec vend starts with two in its stock.

### DIFF
--- a/Resources/Locale/en-US/extenddescriptions/items/weapons/civiliandisabler.ftl
+++ b/Resources/Locale/en-US/extenddescriptions/items/weapons/civiliandisabler.ftl
@@ -1,1 +1,1 @@
-civiliandisabler-extenddescription-security = This weapon is legal to own and carry without a license on NanoTrasen stations, but may be considered as an accessory equivalent to a weapon if used in a crime.
+civiliandisabler-extenddescription-security = This weapon is legal to own and carry without a license for command members on NanoTrasen stations, but may be considered as an accessory equivalent to a weapon if used in a crime.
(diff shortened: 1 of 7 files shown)


Timfa2112 (changes requested):
All command members already have access to normal disablers in their loadout options, and the civilian disabler is expressly made to give other players a weapon that would be effective against spiders and such. It's not intended as a viable weapon against other humanoids.
This change removes the entire point of the civilian disabler's existence, and makes it selectable only to jobs that already have a vastly superior version available to them. I think it would be better to change their stamina damage, or otherwise make them less viable against humanoids, while still being an effective option against spiders or slimes.

LaryNevesPR:
I think the judo belt change is a great change. Removing the Civilian disabler makes no sense, they are extremely weak.
Honestly, all of these points could have been noticed before I even added him to the game and even so, I haven't seen that much valid hunting on my server.

EctoplasmIsGood:
> I think the change to the Judo belt is great. Removing the Civil disabler doesn't make sense, they are extremely weak. Honestly, all of these points could have been noticed before anyone even added it to the game, and even then I didn't see much worthwhile hunting on my server.

They aren't as "extremely weak" as you'd think. They are still capable of stunning, and because they do stam damage, they radically slow down the target before they even stun them. It has led to a large amount of valid hunting on Goob MRP, and I believe some incidents on Deepstation. TCJ, I believe, wants them gone entirely; this is a compromise.

PR #2384, opened 2025-05-03
Plasmemes can now wear full modsuits.

Diggy0:
Yay

# Plasmamen no longer burn when they have a modsuit helmet on.

Literally what it says on the title.

What this PR changes is it adds a type in the YML of the modsuit helmet, that disallows the plasmamen from burning when they equip a helmet piece.

There is a module to fix this issue with plasmamen in 13, so once said modules are implemented we could just add that and revert this.

---

# TODO

- [x] Completed Task

---

<p>

image

Plasmamen with normal modsuit, no longer burns.

</p>
</details>

---

# Changelog

:cl:
- tweak: Plasmamen can now wear Modsuit helmets.

### DIFF
--- a/Resources/Prototypes/_Goobstation/Entities/Clothing/Head/modsuit.yml
+++ b/Resources/Prototypes/_Goobstation/Entities/Clothing/Head/modsuit.yml
@@ -125,4 +125,11 @@
   - type: BatterySelfRecharger
     autoRecharge: true
     autoRechargeRate: 2
+  - type: Tag
+    tags:
+    - PlasmamanSafe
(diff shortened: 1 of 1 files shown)


Diggy0:
The oooonly issue I can think about is that even when unsealed the suit still prevents burning?
But at the moment we don't have code or anything that goes "Only apply these components when sealed" as far as I know.

OldDanceJacket:
Again good for now but likely to have the 13 module that makes them safe for Plas' instead of it just being default.... Probably? Guess we can test what we prefer.

PR #2403, opened 2025-05-04
Birdogical ascendence

Raikyr0:
# Description

This is a test PR to see if one could replace limbs with biological ones instead of cybernetics using the trait.

This is an affront to god.

This trait swaps your arms out for Harpy Wings. There's no way to customize wing color, markings or anything.

# Changelog

:cl:
- add: I profane gods creation and spit in his face, trait that replaces your arms with Harpy Wings.

### DIFF
--- a/Resources/Locale/en-US/traits/traits.ftl
+++ b/Resources/Locale/en-US/traits/traits.ftl
@@ -653,3 +653,9 @@ trait-description-IPCBrittleBoneDisease =
     Made from weaker materials, or corroded by age or damage, your body is much weaker than the standard for an IPC.
     This trait reduces your threshold for death by 60 points.
     (Just as the normal version, it halves the threshold. IPCs cannot be crit, so it instead makes you die 60 points sooner.)
+
+trait-name-Wings = Wings
+trait-description-Wings =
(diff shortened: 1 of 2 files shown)


Erisfiregamer1 (approved):
**further hysterical evil laughter**
If we're going to use the Cybernetic replacement code like that we should probably rename it to be something more generic and make sure it works normally with non-Cybernetic parts, but if it works in testing it's probably fine.

VMSolidus (changes requested):
I should absolutely clarify that yes this is HILARIOUS, it definitely shouldn't be merged. It's a funny proof of concept though that it is indeed possible! When we get around to reworking Markings such that they can utilize CharacterRequirements, then we can actually absolutely revisit this trait!

Raikyr0:
> I should absolutely clarify that yes this is HILARIOUS, it definitely shouldn't be merged. It's a funny proof of concept though that it is indeed possible! When we get around to reworking Markings such that they can utilize CharacterRequirements, then we can actually absolutely revisit this trait!

Can't wait to steal tail dragging from Lizards if you have a tail Marking.

PR #2404, opened 2025-05-04
Magivend Stuff

Raikyr0:
# Description

Allows for the Magivend to be restocked and added pondering orbs to the vend (and RGB staffs)

# Changelog

:cl:
- add: Added fun but actually real


OldDanceJacket (approved):

OldDanceJacket:
For PR's we require media to be attached to the PR(even for something small like this), we'll just handwave it this time just keep it in mind for any future PR's.

PR #2417, opened 2025-05-08
Welder Swords

EctoplasmIsGood:
# Description

Ports welder swords from Goob LRP, they are cheap "energy swords" that do a bit more damage then normal welders and look *really* cool

<p>

image

</p>
</details>

# Changelog

:cl:
- add: Added the welder esword!


89Eblan:
As a WWDP contributor, please dont port this

OldDanceJacket:
> As a WWDP contributor, please dont port this

I have my own thoughts but I'm curious on why you're saying to not port this?

OldDanceJacket (changes requested):
First off you have a seemingly unrelated change in the mix: image

Why?

Secondly the 'cost' of making this is an absolute joke. Oxygen Tank + Wires + Cables. 

Next along is the damage, heat is a resisted damage type and conflicts with slash so this isn't a particularly 'strong' weapon, however it also parents off of the EnergySword so it's attackrate/range will be coming off of that, as would it's power attack. It'd be far smoother to make this have more damage but be a bit more 'unwieldy' with a slower attack rate and give it a sharper direction than _"Cool looking weapon that does mean damage to un-armored fire weak species"._

This is very cool however! So I'm saying give it a sharper mechanical identity, and make it more difficult to construct instead of just outright denying it. 

(I think a lot of people running around with this would negatively impact how a round would look and feel, when it seems like the intention is a crazed engi with a mask running out with a jerryrigged weapon being the theme of the weapon?)

EctoplasmIsGood:
> First off you have a seemingly unrelated change in the mix: image
> 
> Why?
> 
> Secondly the 'cost' of making this is an absolute joke. Oxygen Tank + Wires + Cables.
> 
> Next along is the damage, heat is a resisted damage type and conflicts with slash so this isn't a particularly 'strong' weapon, however it also parents off of the EnergySword so it's attackrate/range will be coming off of that, as would it's power attack. It'd be far smoother to make this have more damage but be a bit more 'unwieldy' with a slower attack rate and give it a sharper direction than _"Cool looking weapon that does mean damage to un-armored fire weak species"._
> 
> This is very cool however! So I'm saying give it a sharper mechanical identity, and make it more difficult to construct instead of just outright denying it.
> 
> (I think a lot of people running around with this would negatively impact how a round would look and feel, when it seems like the intention is a crazed engi with a mask running out with a jerryrigged weapon being the theme of the weapon?)

This weapon is meant to be a jury rigged impersonation of an esword which is why it’s parented. It’s cheap because it’s a pretty shit weapon altogether and you’re basically just turning a normal welder into contra for a tiny stat boost. 

The unrelated change was just cuz of duplicate tags that somehow got added by the cherry-pick, it doesn’t actually effect anything.

EctoplasmIsGood:
> As a WWDP contributor, please dont port this

Okay

PR #2420, opened 2025-05-09
Asterisk: New Center Design & Fixes

mi8hip:
# Description

The central area around the vault has been re-designed to have a more snowy look.

---

<p>

image
image

</p>
</details>

---

# Changelog

:cl: mrs
- tweak: Asterisk: Central area around the vault has been re-designed.
- fix: Asterisk: Sideways dock in the shuttlecrafting area and AI chute being blocked by computers.


OldDanceJacket:
So this center area looks a lot better with this change, but I have a few concerns...

This center area looks nice, the design of it isn't spread to the rest of the station, or even in the immediate area. This is a minor issue, it can't be expected to have it spread everywhere especially when the design is being tested here, but it is worth pointing out.

Next up would be the railing, the rails at first glance seem like a nice touch but then you notice that on the East, West, and North directions you can see that the layering completely falls apart with tables and benches.

There's also the long standing issue with any 'park' area, and that this place is primed for spacing, currently I've presented the plan to put Asterisk back up into space, and even if not this area is concerned with it regardless, the new visual direction is worth keeping though! But how would something like this work you think? 
image


I plan on merging this once tests finish regardless(as it's generally an improvement), but figured I'd get this feedback in because I don't know your username on discord.

OldDanceJacket (approved):

mi8hip:
Sorry for the late reply,

I plan on expanding the visual design to the rest of the station in another PR as the center kind of looks out of place now. I will implement your suggestion since having a large undivided area _usually_ ends up being an atmos nightmare (See literally any other map), even though I don't think it'll be a huge issue as Asterisk rarely has any depressurization.

As for the railings and table/bench, removing them and having park benches on the snow will give the room more of a natural look. Would you say you prefer softer lighting in the center or keeping the LED lights I used to make it look like sunlight?

The original plan was to have an actual "outside" area like in Glacier, but it's not worth making players wear insulated clothing and causing space wind or placing tiny fans on every door in a SINGLE high traffic room just to play to the map's gimmick.

I've also had some issues with the map not saving after multiple major changes which was very upsetting and took some time to fix. 

My discord username is @mi8hip if you wish to contact me.

PR #2426, opened 2025-05-15
Machine Languages

VMSolidus:
# Description

This PR fixes a bug whereby the Holopads weren't respecting languages, and actually basically every machine in the entire game wasn't. There's an in-general broader issue that I would reaaaaaally like if UI elements in general could be differentiated by language, but that's a lot harder to do than this fix. This was shockingly easy to fix actually.

# Changelog

:cl:
- fix: Holopads now correctly respect the speaker's language, and transmit said language to the receiver. They are no longer Universal translators. Have fun yelling at people over the holopad in whatever obscure language your character knows.
- fix: Nearly every machine that can speak, now correctly respects that languages exist. For example, mice can now no longer understand what vending machines are saying.


Erisfiregamer1 (approved):
Yeah. I don't think more has to be said.

PR #2448, opened 2025-05-23
Cyberware Code Support

VMSolidus:
# Description

Another request by Aisha for Hullrot. This adds the necessary code for supporting completely yml serialized cybernetics, via use of the all powerful TraitFunctions library. A new TraitFunction exists for removing actions from an entity, and in order to make it less terrible to use, I added a new public function to the SharedActionsSystem to allow removing an Action from an entity via Action Prototype rather than EntityUid. 

# Changelog

:cl:
- add: Added support for Organs(such as Cybernetics) to utilize the same massive library of functions used by Traits.

### DIFF
--- a/Content.Server/Traits/TraitSystem.Functions.cs
+++ b/Content.Server/Traits/TraitSystem.Functions.cs
@@ -129,6 +129,23 @@ public sealed partial class TraitAddActions : TraitFunction
     }
 }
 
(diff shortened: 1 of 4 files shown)


Erisfiregamer1 (approved):
Yeah, nobody's gonna care if I just...

PR #2449, opened 2025-05-23
Wizmerge

Spatison:
# Description

Please don't squash it.

---

# TODO

- [x] Porting all fixes
- [x] Make sure it works

---

# Changelog

:cl:
- add: A large amount of content has been migrated from Wizden


Erisfiregamer1 (changes requested):
Didn't we say we were going to stop doing Wizmerges because they kept overwriting stuff we patched in, like Contests?

This needs much more extensive review for approval. We don't need a repeat of the issues the last Wizmerges caused.

Spatison:
So, this is ready for review, but I need to port the arena admin fix first

VMSolidus:
**VERY CAREFULLY MAKES SURE NOT TO PRESS SQUASH**

PR #2471, opened 2025-05-31
More Drain System Perf Tweaks

VMSolidus:
# Description

Actually the bulk of this system's cost was in having these expensive entity querries running on every frame, so instead of interrupting the drain logic on a per drain basis, we interrupt the entire system of querries itself globally, and cut off the main expense of the system at its root source.

# Changelog

:cl:
- tweak: Made some significant performance improvements to Drains.


bruhmogus:
does this make them actually work too

PR #2472, opened 2025-05-31
Gas Tile Overlay Performance Tweaks

VMSolidus:
# Description

Gas Tile Overlay had a non-functional check for preventing it from running on every frame. This PR fixes the frametime stopper, such that it actually prevents the system from constantly running at all times.

# Changelog

:cl:
- tweak: Made some performance improvements to the Gas Tile Overlay.


Erisfiregamer1 (approved):
we should add a tray but for atmos, like it shows the vector of force of the atmospheric movement around you

PR #2483, opened 2025-06-11
Remove buttons for Medical Examination, Detail description, and ID inspection.

Stonetear:
# Description
Removes the annoying verbs for inspecting someone's character description or health status.

Screenshot from 2025-06-10 18-47-26

Also renames file `DetailExaminableystem.cs` to `DetailExaminableSystem.cs`.

---

# TODO
[ ] Integrate ID inspection into the Inspection tooltip
[ ] Consider changing detail description components to be Shared instead of Server, which would remove the delay.
[ ] Change character detail descriptions to display the first 3 lines of a character description, then also have the button for particularly long-winded descriptions on their own page.

---

# Changelog

:cl: Stonetear
- tweak: Medical examination, ID examination, and Detail examination no longer require additional clicks.


Erisfiregamer1 (changes requested):
If you're going to do this then it'd be worthwhile to move it to Shared.

Also, can you make it so that the old behaviour is still available for the future? I *would* like to be able to keep adding buttons to the description for longer additions.

Erisfiregamer1:
> > If you're going to do this then it'd be worthwhile to move it to Shared.
> > Also, can you make it so that the old behaviour is still available for the future? I _would_ like to be able to keep adding buttons to the description for longer additions.
> 
> This doesn't belong in shared. Sending ID name is done by the server regardless(since it only has the data) Not to mention it might cause race condition on any other system that is reliant on server-data

I see then.

WhiteHusky:
Might want to add a hint that the player should move closer for a detailed examination in some small text in the examination (if the component is there.)

PR #2485, opened 2025-06-12
Fix Mailing Units (space-wizards/space-station-14#30174 + Other)

WhiteHusky:
# Description

_Upstream: space-wizards/space-station-14#30174_

This fixes mailing units so their UI works again alongside their sprites. It also adds a convenience feature to reset the destination on send since it can trip up players as it makes sense to put in an item _then_ choose a location.

# Changelog

:cl:
- fix: Mailing Unit UI is now fixed
- fix: Mailing Units' no longer use the wrong sprite when charging
- tweak: Mailing Units clear target on send or selecting the same target


WhiteHusky:
On the by, this was downstream'd on Fansana/floofstation1#831 and, anecdotally by me, has fixed the units so they can be useful again for inter-department mailing with units.

PR #2488, opened 2025-06-17
Psionic Refactor V4

VMSolidus:
# Description

This PR refactors Psionics code in order to open it up to significantly more modularity, by making extensive use of some of the strongpoints of ECS code. The general gist of what I'm trying to do is make it so that the Psionic Component generally only acts in its capacity as a mediator for these systems to all have a consistent basis to interact with each other, and should not contain logic for modifying itself- rather it should include a hierarchy of events and subscriptions that give a consistent order of operations which other components can hook into and make their needed modifications to the PsionicComponent. 

The important breaking changes are that Potentia, Glimmer, and the new Neural Heat mechanic are all separate components from the Psionic Component, and may be included or omitted as-desired by simply adding the components needed. These three systems would themselves also have an event-subscription model, where the actual logic is handled by further smaller components which only perform a single task.

Neural Heat is being introduced as a new mechanic that allows for more personal consequences to be added for psicasting, although in this PR it will largely be unused outside of the basic implementation to establish its required math.

# TODO

- [ ] New handling for "Psi-Caster Prototypes", which are a way of quickly assembling a desired psionic user from premade yml lists. 
- [ ] Separate Potentia into its own system-component hierarchy.
- [ ] Separate out Glimmer generation into its own system-component hierarchy
- [ ] Neural Heat mechanic as its own system-component hierarchy.
- [ ] Rework psionic powers themselves to use new event-subscription for letting components be responsible for interacting with the powers in a generalizable manner.
- [x] Refactor innate powers to use the new init ordering. 
- [x] Add discreet event and subscription hierarchy for Psionic Component initialization.

<p>

Example Media Embed

</p>
</details>

# Changelog

Fill the changelog later.
:cl:
- add: Added fun :D
- tweak: Tweaked fun
- fix: Fixed fun!
- remove: Removed fun :(


a-person5660:
psionic refactor 4: electric boogaloo

PR #2502, opened 2025-07-11
Fix Traits Anticheat

VMSolidus:
# Description

Traits Anticheat was not accounting for traits having variable slot occupancy, and was treating the "0 slot traits" as if they always had a slot cost of 1. This PR corrects this by making it count the actual slot costs of traits for the purpose of checking for illegal totals.

# Changelog

:cl:
- fix: Fixed a bug with Traits Anticheat incorrectly triggering if the player had selected enough 0 slot traits.


Erisfiregamer1 (approved):
Speedmerge this shit immediately

PR #2506, opened 2025-07-17
Fix SpaceWindv5 Wall Smoothing

VMSolidus:
# Description

There was a math error in the "wall smoothing" step of space wind hiding in plain sight which was causing the "Wall Bugs", this PR fixes that. Behold in all its video proofy glory.

https://github.com/user-attachments/assets/91e87324-b3a9-4d08-bca7-365b64c4888c

# Changelog

:cl:
- fix: Finally completed Space Wind V5's airflow math. It no longer produces extremely inconsistent airflow values in and around airlocks and walls.


Erisfiregamer1 (approved):
:+1:

PR #2513, opened 2025-07-24
better magboots sprites

Ghost581X:
now they shouldn't stick out awkwardly while in hardsuits
port from hurrot

image


Erisfiregamer1 (approved):
Wait, that's an unrelated fail. What am I doing?

PR #2526, opened 2025-08-06
Psionic overhaul

bunnygirlHyacinth:
# Description

added new toys for the justicar, a special pair of gloves and a mask. this is also going to be used for the sorcerer antag eventually

---

# Changelog

:cl:
- add: enchanter gloves
- add: sorcerer mask ( real and fake versions )
- add: both items available in justicar loadout for a total of 15 points all together


bunnygirlHyacinth:
SORRY WRONG BUTTON

PR #2527, opened 2025-08-06
Mollytesting

starmynx:
# Description

Copied the drinks.ftl file from its original location in prototypes to a new location in the en/us section of the locale folder. the originally misplaced file remains in case it is needed in its original location and because I already made the commit so here we are bleh.

# Changelog

:cl:
- tweak: copied the name and description file for the vodka soyjack to a new location to fix it not working.


Erisfiregamer1:
The everpresent fault of pressing the wrong god damn button when PRing your changes...

PR #2530, opened 2025-08-08
Anchor v1.3

OldDanceJacket:
# Anchor fixes

Anchor had a number of smaller issues I can't even remember at this point. We went through and fixed them.

Oh also Cargo got reworked to be bigger because it was too small.

---

# Changelog

:cl: ODJ
- fix: Anchor station bugs fixed


astriloqua:
Test fail seems relevant. Did you remove the boxer spawn points but forget to remove the boxer from the map prototype?

